PR #199, opened 2020-01-08
HTML and accessibility tweaks

zooley:
Admittedly, I feel somewhat squeamish about a PR with a lot of commits but all of these were relatively simple changes and related to the scans/checks I was performing. I can break these up into individual PRs if desired.


aebruno:
@zooley Apologies for the delay. These are all great, thanks so much for the PR. @mkzia and I are reviewing and will hopefully get this merged in shortly.  No need to break these up.

aebruno:
Thank you!

PR #200, opened 2020-01-14
Fix deletion of invoice notes

zooley:
A couple minor HTML fixes and two changes that were preventing deletion of invoice notes.

### DIFF
--- a/coldfront/core/allocation/forms.py
+++ b/coldfront/core/allocation/forms.py
@@ -159,7 +159,7 @@ class AllocationReviewUserForm(forms.Form):
 
 class AllocationInvoiceNoteDeleteForm(forms.Form):
     pk = forms.IntegerField(required=False, disabled=True)
-    note = forms.CharField(max_length=64, disabled=True)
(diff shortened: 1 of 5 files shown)


dsajdak:
@zooley I'm not able to replicate the issue with deleting the invoice notes, unless you're talking about after the allocation is in 'paid' status.  Can you just let me know your process so I can try to replicate before we merge this PR?

zooley:
How long are your notes? All the issues popped up due to the form having `max_length=64` for the `note` field. If I added a note longer than 64 characters, which is allowable as the database column is a TEXT field, and then tried to delete the note from the invoice page, the `if formset.is_valid()` check would always return false (due to the `note` being longer than 64 characters). This, in turn, would cause an error to get thrown because the `return HttpResponseRedirect(...)` line is nested under that `if` statement and the method would return `null` (and the method requires an `HttpResponse` is returned).

dsajdak:
Ah, yeah I missed that part.  Yep, I can replicate the problem.  Thanks for the fix!

aebruno:
Thank you!

PR #205, opened 2020-02-11
Default to '[no journal info from DOI]' when journal info is absent

artlogic:

This is a fix for #204 where some DOI entries are lacking a journal field.


aebruno:
Thank you!

PR #208, opened 2020-02-21
Add missing auth mixins to GrantCreateView

artlogic:

The test_func method was defined, but without the UserPassesTestMixin,
Djano wasn't utilizing it.

As a result, any user could create a grant for any project, by using the
proper url (with the project primary key in the appropriate place).

Additionally, the LoginRequiredMixin was absent, allowing anyone to make
such a change.


aebruno:
Thanks for catching this.

PR #209, opened 2020-02-21
Add a ResearchOutput type freeform

artlogic:

Add a freeform form for users to report non-publication data generated on the HPCC.


dsajdak:
@artlogic could you please provide more information on how this would be used at your center?  It looks interesting to us but we think it might be better placed in a plug-in that other users of Coldfront could enable only if they want to use this feature.  We'd like to hear more about the use case though.

keenandr:
@dsajdak There are reportable outcomes of our researchers' projects that are not publications that we want to collect: open source software released, tech transfers, education efforts, community outreach, conference presentations, preprints, etc. Collecting this data lets us present the full picture of the results of our researchers' work.

aebruno:
Thanks for the PR. Here's some initial feedback:

Getting an error with the migrations when using MySQL:

```
_mysql_exceptions.OperationalError: (1170, "BLOB/TEXT column 'description' used in key specification without a key length")
django.db.utils.OperationalError: (1170, "BLOB/TEXT column 'description' used in key specification without a key length")
```

Looks like a length is required for the description column as it's being used in the constraint:

```python
migrations.AddConstraint(
    model_name='researchoutput',
    constraint=models.UniqueConstraint(fields=('project', 'title', 'description'), name='unique_entry_in_project'),
),
```

Perhaps we should re-think how we define uniqueness here. Maybe just simply use the ID at the db level and add some application logic for checking duplicates? The project, title, desc constraint doesn't seem very robust.

dsajdak:
@keenandr this is really interesting and something our director is definitely interested in incorporating into ColdFront.  The only thing we're concerned with is the freeform text and the likelihood of getting garbage (or less formally formatted) data.  Have you considered adding a category to this to attempt to capture at least some of the potential types of output?  For example, categories like: preprints, software products, tech transfer (patents, etc), outreach (tours, events, presentations), and then the generic "other" category.  He is also getting asked at CASC meetings if ColdFront could be extended to have PIs report on masters and PhD degrees granted annually so this would be a good place for that.

artlogic:
@aebruno I'm looking into other solutions for this problem. The constraint only exists to protect against the case of an accidental double submission, and I don't imagine that's incredibly likely anyway. For the moment I'm fine with removing it and revisiting the problem later if that's acceptable to you and @keenandr.

aebruno:
@artlogic Sounds good. Let's remove the db constraint until we come up with a better solution. Might make sense to add some application logic to check for dups? 

Also, see @dsajdak comment regarding adding categories. Would be great to add the following categories as a start and make this a requirement when submitting:

- preprints
- software products
- tech transfer (patents, etc)
- outreach (tours, events, presentations)
- generic "other" category

Thoughts?

keenandr:
@artlogic +1 on removing the constraint.

@aebruno I think @dsajdak 's ideas for categories are a good idea. I'll discuss implementation details with @artlogic

aebruno:
@keenandr All sounds great. Thanks!

artlogic:
@aebruno The unique constraint has been removed. This should be good for merge. Let me know if you notice any other problems.

Issue #207, opened 2020-02-12
Django 2.2 requires SQLite 3.8.3 or later on CentOS 7


Quoding:
It seems that the Django version required by ColdFront isn't in line with the present version of SQLite on CentOS 7. 

The latest SQLite version yum (with epel enabled) gives is 3.7.17, which is lower than the minimum required by Django 2.2.

It seems that the plausible workaround for now is to install SQLite from source.

This is the output during the initial setup:
```Traceback (most recent call last):                                                                                                                                                   [294/524]
  File "/centos/coldfront_app/venv/bin/coldfront", line 11, in <module>                                                                                                                       
    load_entry_point('coldfront', 'console_scripts', 'coldfront')()                                                                                                                           
  File "/centos/coldfront_app/coldfront/coldfront/__init__.py", line 11, in manage                                                                                                            
    execute_from_command_line(sys.argv)                                                                                                                                                       
  File "/centos/coldfront_app/venv/lib64/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line                                                  
    utility.execute()                                                                                                                                                                         
  File "/centos/coldfront_app/venv/lib64/python3.6/site-packages/django/core/management/__init__.py", line 357, in execute                                                                    
    django.setup()                                                                                                                                                                            
  File "/centos/coldfront_app/venv/lib64/python3.6/site-packages/django/__init__.py", line 24, in setup                                                                                       
    apps.populate(settings.INSTALLED_APPS)                                                                                                                                                    
  File "/centos/coldfront_app/venv/lib64/python3.6/site-packages/django/apps/registry.py", line 114, in populate
    app_config.import_models()
  File "/centos/coldfront_app/venv/lib64/python3.6/site-packages/django/apps/config.py", line 211, in import_models
    self.models_module = import_module(models_module_name)                                     
  File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module                
    return _bootstrap._gcd_import(name[level:], package, level)                                
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import                               
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load                            
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked                   
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked                            
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module                      
[...]

PR #211, opened 2020-03-23
Update README.md

Quoding:
Update minimum CentOS version to 8 since it is the minimum requirement for the Django version used by ColdFront.


aebruno:
Thanks for the PR. Any chance we can describe this in a bit more detail? Would love to see use document any work around we can provide for centos7 users.

Quoding:
I have updated the README with further information as to how to install SQLite 3 from source for CentOS 7 users. This was my workaround as well as another user's in the issue I opened (#207). I made two sections, one for centos 7 and another for centos 8.

aebruno:
Thanks!

PR #215, opened 2020-05-26
List who is responsible (PI, managers) for a given user

artlogic:
_This is a part of a number of PRs that I'm currently working on for @keenandr from MSU._

This PR provides a form gated with is_staff that allows the user to input a username, search matching projects, display project and managers for that project.

### DIFF
--- a/coldfront/core/user/templates/user/user_profile.html
+++ b/coldfront/core/user/templates/user/user_profile.html
@@ -2,7 +2,7 @@
 {% load staticfiles %}
 
 {% block title %}
-User Profile
+User Profile{% if not user == viewed_user %}: {{ viewed_user.username }}{% endif %}
 {% endblock %}
 
 {% block content %}
@@ -14,9 +14,11 @@ User Profile
 
 
   <div class="card ">
-    <div class="card-header"><i class="fas fa-user" aria-hidden="true"></i> User Profile</div>
+    <div class="card-header"><i class="fas fa-user" aria-hidden="true"></i>
+      User Profile{% if not user == viewed_user %}: {{ viewed_user.username }}{% endif %}
+    </div>
     <div class="card-body">
-      <h4 class="card-title text-center">{{user.first_name}} {{user.last_name}}</h4>
+      <h4 class="card-title text-center">{{viewed_user.first_name}} {{viewed_user.last_name}}</h4>
       <div class="table-responsive">
     <table class="table">
       <tbody>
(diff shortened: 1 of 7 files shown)


dsajdak:
@artlogic We are missing a step with this feature.  When I give someone "staff status" they are not able to see the "Admin" menu where the "user search" feature is.  If I navigate to the direct URL for the user search page, the staff member can use it and see all of this new profile information.  However, we need the User Search on a menu somewhere so they can access it.  Since non-admins don't have access to the admin menu, where are you displaying this menu option in your deployment?

Otherwise, the functionality of this all is great!

artlogic:
@dsajdak thank you for the feedback! I'm looking into this omission and should have a patch ready in a few days.

aebruno:
> I'm looking into this omission and should have a patch ready in a few days.

@artlogic Any update? Don't mean to rush but we're gearing up for a new release and would love to have this included. Thanks again for your contributions!

artlogic:
@aebruno I'll be working on this in the next 24 hours. Thanks for checking in!

artlogic:
> @artlogic We are missing a step with this feature. When I give someone "staff status" they are not able to see the "Admin" menu where the "user search" feature is. If I navigate to the direct URL for the user search page, the staff member can use it and see all of this new profile information. However, we need the User Search on a menu somewhere so they can access it. Since non-admins don't have access to the admin menu, where are you displaying this menu option in your deployment?
> 
> Otherwise, the functionality of this all is great!

@dsajdak @aebruno Just spent a bit of time digging into this and when I grant a user staff status, a "Staff" menu is displayed with the "User Search" option. Can you tell me more about how you are granting someone staff status?

dsajdak:
@artlogic we are not seeing that "Staff" menu.  We make users "staff" by checking the "Staff Status" checkbox under their user account in django.

staff

Even if we check the "superuser status" checkbox, this Staff menu doesn't display.

artlogic:
Here's the process I followed to see the "Staff" menu:

1. Start a fresh instance.
2. Run initial_setup and load_test_data.
3. Login in as admin.
4. In user administration, go to users, and then click on cgray. Click the "Staff status" checkbox and then save.
5. Logout. Then Login as cgray.
6. Here's the menu I see:

Screenshot from 2020-07-13 13-49-15

Here's the code that causes that menu to appear:

https://github.com/ubccr/coldfront/blob/654694ed93c3cdf753ee4ed667e5668ae67e106b/coldfront/templates/common/authorized_navbar.html#L32-L33

And here's the menu itself:

https://github.com/ubccr/coldfront/blob/654694ed93c3cdf753ee4ed667e5668ae67e106b/coldfront/templates/common/navbar_nonadmin_staff.html

My thought right now is that something in your config is causing `request.user.is_staff` to not work. Could you possibly emit this value to see what it is? Also could you verify that if you change the condition to be explicitly true that you see the menu?

dsajdak:
@artlogic very sorry about that!  We have custom templates we use for our branding and needed to import your changes.  This looks great!

aebruno:
@artlogic Thanks again for the great work.

Issue #34, opened 2018-09-06
Add export feature for publications


dsajdak:
All users to export their publication list

dsajdak:
Export to bibtex format.  Allow user to select some or all of publications to export

PR #216, opened 2020-06-04
Export feature

bhatiadheeraj:


dsajdak:
I verified this works in the test instance.  Can export or delete one or many publications

aebruno:
This PR fixes #34. @bhatiadheeraj be sure to include the issue that your PR will be addressing. Thanks!

PR #219, opened 2020-06-23
Username search fix

bhatiadheeraj:


aebruno:
@bhatiadheeraj This PR also includes all the commits for the export feature, which is part of PR #216. Please close this out, and rebranch off master so that your commit includes only the fix for user search.

PR #234, opened 2020-08-25
Accessibility fixes, minor HTML fixes

zooley:
This makes the following changes:

* Add the `aria-hidden` attribute to decorative icons that were missing it.
* Add `scope` attribute to table headers that were missing it.
* Add text for elements that only had icons--Note this keeps the text hidden to all but screen readers, so there is no visual change for standard users.
* Minor HTML fixes.

Other issues:

* It's generally recommended for tables to have the `caption` element. Some accessibility scanners didn't complain (Deque's aXe) while some will flag it as an issue (ComplianceSheriff). One could add the element but hide from standard view (`<caption class="sr-only">Text</caption>`) to avoid any visual changes. Considering every table has some form of header preceding it, I would think the repetition of text could be annoying to those using screen readers. As such, I'm a little uncertain on this one and will do some more research before proceeding.
* Incorrect heading order. The fixes for this may involve further changes to the primary CSS to maintain styles. This may be addressed in a separate PR.
* There are a number of color contrast accessibility issues. I had started work on this several months ago but time got away from me. I will see about submitting some of that as a separate PR.

Ref: https://webaim.org/techniques/tables/data
Ref: https://dequeuniversity.com/rules/axe/3.5/heading-order

### DIFF
--- a/coldfront/core/allocation/templates/allocation/allocation_add_users.html
+++ b/coldfront/core/allocation/templates/allocation/allocation_add_users.html
@@ -57,7 +57,7 @@ Add Users to Allocation
   </div>
 </div>
 {% else %}
-<a class="btn btn-secondary mb-3" href="{% url 'allocation-detail' allocation.pk %}" role="button"><i class="fas fa-long-arrow-left"></i> Back to Allocation</a>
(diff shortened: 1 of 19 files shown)


aebruno:
This looks great. Thanks so much!

Issue #248, opened 2020-11-18
Publications: allow users to see full list of publications on the project


dsajdak:
Add an option in this drop down list to include 'all'
![pubs](https://user-images.githubusercontent.com/9355962/99566426-e21a2a00-299a-11eb-8992-a35ad94b360a.PNG)

PR #250, opened 2020-11-18
View all publications option from project detail page - issue #248

dylan-tkt:
re: Issue #248 

Screenshot 2020-11-18 at 17 24 04

looks like publications query in context builder is not limited, so the user should now have access to all publications if they select this option via dropdown.

Line 123-124: core/project/views.py

```python
context['publications'] = Publication.objects.filter(
    project=self.object, status='Active').order_by('-year')
```


dsajdak:
Thank you so much for the contribution!

PR #265, opened 2021-03-04
Skip disabled resource.

### LINKED ISSUE #235: Slurm Plug-in:  disabled resource should not show up in slurm files
If a resource is unmarked 'available' in the database, the coldfront slurm dump should not include that resource and it's attributes.

aebruno:
Fixes #235. Disabled resource should not show up in slurm dump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # ColdFront Changelog
 
+## [Unreleased]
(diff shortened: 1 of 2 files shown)


dsajdak:
Looks good.  Works in dev environment

dsajdak (approved):

PR #266, opened 2021-03-04
Fix year validation bug.

### LINKED ISSUE #258: Publications: manual entry data validation issue
When entering a publication manually, if a non-date is entered in the date field an error is thrown:

publicationerror

aebruno:
This fixes #258. Add better form validation for manual publication entry. Change to using uuid4 instead of sha256 hash for unique ids. The issue is when manually entering a publication, we do not ask for a DOI as it may not have one. As a doi uniquely identifies a publication, we simply use a randomly generated uuid4 as the unique id.

### DIFF
--- a/coldfront/core/publication/views.py
+++ b/coldfront/core/publication/views.py
@@ -1,7 +1,5 @@
-import hashlib
-import json
 import re
-
+import uuid
 import requests
 import os
 import io
@@ -16,6 +14,7 @@ from django.http import HttpResponseRedirect, HttpResponse
 from django.shortcuts import get_object_or_404, render
 from django.urls import reverse
 from django.views.generic import DetailView, ListView, TemplateView, View
+from django.views.generic.edit import FormView
 from django.views.static import serve
 
 from coldfront.core.project.models import Project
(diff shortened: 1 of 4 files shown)


coldfront/core/publication/views.py
```diff
@@ -247,80 +246,61 @@ def post(self, request, *args, **kwargs):
...
+        project_obj = get_object_or_404(Project, pk=self.kwargs.get('project_pk'))
 
         if project_obj.pi == self.request.user:
             return True
 
         if project_obj.projectuser_set.filter(user=self.request.user, role__name='Manager', status__name='Active').exists():
             return True
 
+        messages.error(self.request, 'You do not have permission to add a new publication to this project.')
+
```
dsajdak, line 267: I see this error in the django admin interface, if I'm using the 'su' option to login as the user.  However, if I login directly as the user, this results in a 403 error with no explanation.

dsajdak (changes requested):
Works in dev environment.  Suggested one fix regarding user role login.  There is no way for the user to get this error without going directly to the URL created when adding a publication as there are no 'add publication' buttons show to the user.  So this could be skipped.

coldfront/core/publication/views.py
aebruno: Added 403 template so that those errors are now displayed.

dsajdak (approved):

Issue #259, opened 2021-02-24
Grants:  Archived grant not viewable by PI


dsajdak:
When a PI changes the status of a grant to 'archived' they can no longer see it on the project detail page.  There is no grant page to list all of them.  If they try to delete grants, all of them on the project are visible.  It would be nice for the PIs to be able to see the status of the grant and be able to toggle the view depending on status.

PR #299, opened 2021-06-02
Archived grants can be viewed on the Project Detail page

brisco17:
Resolves issue #259.  The grants section of the page also has a new status column to show the status of each grant associated with a project.


dsajdak (approved):
This looks great!

PR #300, opened 2021-06-02
Multiple Allocations on a project for the same resource can be differentiated

### LINKED ISSUE #193: Allocations: multiple allocations on a project for same resource need more info
If there are multiple allocations on a project for the same resource, when a user gets added to the project there is no way to tell the allocations apart.  We need to add an slurm/cloud account name or unix group attribute as part of the allocation name.

brisco17:
Resolves #193. In the allocation section of a project, the information column for an entry will display its slurm_account_name, freeipa_group name, Cloud Account Name and/or its Cloud Storage Quota. If an entry does not have any of these attributes, its description will be used to fill out the information column instead.

### DIFF
--- a/coldfront/config/core.py
+++ b/coldfront/config/core.py
@@ -41,6 +41,11 @@ SETTINGS_EXPORT += [
 ADMIN_COMMENTS_SHOW_EMPTY = ENV.bool('ADMIN_COMMENTS_SHOW_EMPTY', default=True)
 
 #------------------------------------------------------------------------------
+# List of Allocation Attributes to display on view page
+#------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


dsajdak (commented):
There are 2 parts to this:
1. When a PI adds a user to a project, if they have multiple allocations for the same resource, they can't be differentiated between.  We need to add an attribute here so they can tell which is which.  For cluster or cluster partition resources, display 'slurm account name.'  For cloud resource, display 'cloud account name.'  For storage resources, display 'storage directory name'.  Currently, if 'description' is on the allocation, this is being displayed.
2. On the project detail page, we should display in the 'information' column one or more of these attributes, in order to differentiate.  This is mostly done.  However, the 'project storage' resource should display 'storage directory name' also (or instead of, unix group)

dsajdak (approved):

Issue #203, opened 2020-01-29
Allocations:  provide way to not allow renewing an allocation


dsajdak:
Some allocations can not can be renewed.  We need to provide a way for admins to prevent the renewal of allocations ('lock' doesn't do this).  Emails about upcoming expiration should be different than those that can be renewed.  They should be notified they're about to lose access to the resource but told they can't renew it.

dsajdak:
This should be two-fold:
1. There should be a way to turn off expiration notifications for all users on an allocation with one change (checkbox or something like that)
2. We need to ensure the PI can't renew allocations if we have them marked as a non-renewable allocation

dsajdak:
"renew now" banner should indicate number of days until expiration but show - not renewable, rather than 'click to renew'  Also the banner should not be clickable or redirect to renewal page.

dsajdak:
Current allocation attribute is "EXPIRE NOTIFICATION"

PR #307, opened 2021-06-17
Prevent allocation renewal setting

brisco17:
Resolves issue #203. When an allocation is locked in the ColdFront Administration Portal by an admin, the allocation can no longer be renewed. When an allocation is locked, the "renew now" banners change to only show the number of days an allocation has until it expires. It says "not renewable" instead of "renew now", and is no longer clickable.


dsajdak (approved):
Looks good!  OK to merge

PR #310, opened 2021-07-09
Update docs with correct username of PI#1 and center director

Iain-S:
Hello

I think ccollins and michardson must have been from an older version so I have updated the docs to reflect the users currently in [load_test_data.py](https://github.com/Iain-S/coldfront/blob/7130334c96b263c3c48b6844b6dad6e6b811c827/coldfront/core/utils/management/commands/load_test_data.py)

### DIFF
--- a/docs/pages/install.md
+++ b/docs/pages/install.md
@@ -128,8 +128,8 @@ $ coldfront load_test_data
 ```
 
 - You can log in as `admin` with password `test1234`.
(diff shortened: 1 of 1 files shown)


dsajdak (approved):

dsajdak:
@aebruno this is good to merge.  @Iain-S Thank you for the contribution!

PR #311, opened 2021-07-13
allow logout redirect URL to be configured

culbert:
I need to be able to configure the LOGOUT_REDIRECT_URL to perform OIDC IdP logout.

Note that the default behavior of LogoutView.as_view() is to use the LOGOUT_REDIRECT_URL (thus the change from specifying the redirect to using the default in the user/urls.py causes the redirect to use the LOGOUT_REDIRECT_URL value).


aebruno:
This is great. Thanks so much.

PR #312, opened 2021-07-19
Update docs and config with default Postgres port number

Iain-S:
Even thought the Django-environ [docs](https://django-environ.readthedocs.io/en/latest/) use 8458, the default PostgreSQL port number is 5432. Using this in the coldfront docs may avoid confusion.


aebruno:
Thanks for catching this. Totally agree.

PR #315, opened 2021-07-20
Add ColdFront command to dump user information for import into XDMoD.

jpwhite4:
This adds a `coldfront xdmod_dump` command that outputs user information in json format. Note that XDMoD 9.5 has no mechanism to import this data. I plan to travel to the future to ensure that XDMoD 10.0 can read this file format and set the
names and PI status appropriately.

The command-line name and arguments look-and-feel were chosen to be the same
as the `coldfront slurm_dump` command from the slurm plugin. I assume that this is the informal standard.


aebruno:
@jpwhite4 Thanks for the PR. This looks good however, there's nothing really xdmod specific here. It might make more sense if this were a core user command. i.e. move this to `core/user/management/commands` and call the command `user_dump`. That would make it slightly more user friendly in case other folks have a need for dumping the user information. We also aim to build a proper ColdFront API in the near future and this might be a good use case for that.

PR #317, opened 2021-07-23
environ import (plus some pep8)

Iain-S:
I should probably have done this as 2 PRs:

See 9a5b1d2 for the main change (setting COLDFRONT_CONFIG causes an error because `environ` had not been imported into settings.py).

I then ran flake8 on the config files and fixed most of the errors (I did not address line-length errors or any files under `config/plugins/`).


aebruno:
Thanks!

PR #324, opened 2021-10-01
Attribute expansion

payerle:
This extends the Attribute system to allow for attributes to refer to other attributes and do simple operations with them. As a simple example, the load_test_data has been extended to add a new cluster 'University Metered HPC' which defined the 'slurm_specs' attribute to set GrpTRESMin to restrict the amount of CPU time which can be used by an allocation based on the 'Core Usage' attribute of the allocation.

No schema changes, but a new AttributeType 'Attribute Expanded Text' has been added (under models.py for both allocation and resource). To both AllocationAttribute and ResourceAttribute the new methods typed_value() and expanded_value() have been added; the former will return attributes of type Int or Float as numeric vs string types in python, and the latter returns the value after expanding any referenced attributes in the value. The get_attribute() and get_attribute_list() methods now by default wil expand and "type" their returned values, but also have some optional parameters added to control such behavior. The ResourceAttribute versions also will take an extra_allocations list (defaults to empty list) to allow for some AllocationAttributes to be used when expanding ResourceAttribute. Most of the real work is done in core/attribute_expansion.py

The scheme was desired for modifying 'slurm_specs' according to AllocationAttributes as shown int he test data, but the scheme is fairly generic. Some small changes were made to the slurm plugin to pass the allocation be acted on to the get_attribute_list for the Resources the allocation depends on (in order to allow referencing AllocationAttributes from ResourceAttributes).
The expansion basically works as follows: if an attribute 'foo' is not of the type 'Attribute Expanded Text' type, we just return the unexpanded value (possibly converted to a numeric type in python). If it is of that type, we look for an attribute named 'foo_attriblist'. If that attribute is not found, we again return the unexpanded value. If it is found, each line in foo_attriblist is processed in order to generate a Python dictionary. This dictionary is then given as arguments to the format() method of the Python string unexpanded value of foo to expand it. If there is an error in expanding it, the unexpanded value is returned.

The attriblist has a simple language, consisting of lines of the form:
[keyname] [op]= [argument] 
where
[keyname] is the name of a key in the dictionary to act upon.
[op] is a single character which indicates what operation using [key] and [argument] to perform
[argument] is the argument which along [key] is operated upon.  It can be a string literal enclosed in single quotes (quite simplistically; lead/trailing single quote stripped off and everything else is the string literal), a number, or ':' followed by the name of a key already in the dictionary, or an Allocation- or ResourceAttribute.

[op]= can be
: = for assignment (i.e. value of [argument] is put in the dictionary under [key] )
|= for defaulting (i.e. assign [argument] to [key] only if  value of [key] is None)
+=, -=, *=, /= set [key] to [key]+[argument] , [key]-[argument] , etc.
(= In this case, [argument] should be the name of an unary function, which is applied to the value of [key]and then the resulting value stored in [key]. Currently only recognized function is 'floor'.

I have not made any changes in documentation, but assuming this pull rquest is approved I intend to follow up with the appropriate doc changes.


aebruno:
@payerle Thanks so much for the contribution. This looks great. We're going to test this out shortly on our internal systems and hopefully get this merged soon. Thanks again!

aebruno:
@payerle Apologies for the delay on this. We wanted to finish up PR #330 first and it took a bit longer than anticipated. We're testing this out now and will report back with any feedback. Thanks again for contributing!

dsajdak (approved):
Tested in our dev instance and all looks good.  Also did a fresh install to test out the example data and see how the slurm plugin works.  No issues.

PR #328, opened 2021-10-19
add plugin

claire-peters:
(Mistakenly pushed merge request further upstream than intended, please ignore)


dsajdak:
@geistling We would love to get this upstream whenever you're ready!

Issue #294, opened 2021-05-05
Allow PI to request changes to resource allocation attributes


knikolla:
As a PI, I would like the possibility of requesting a change to resource allocation attributes after a resource allocation has been created and approved. This request would require approval from administrators before being executed.

Resource allocation attributes are used to express quotas/limitations of a resource allocation on a resource, and in a cloud environment where some resource allocations can be very long lived, should be allowed to scale up as usage and requirements grow.

dsajdak:
Possibly related to or in place of #287

dsajdak:
This is a high level diagram of what we'd want the process to look like.  We want to allow the PI/manager to make change requests on SOME allocation attributes, as well as the end date, and possible other things.  This sketches out two potential ways to accomplish this:

![allocation_change](https://user-images.githubusercontent.com/9355962/121422055-cf6a4100-c93c-11eb-895f-af57a58ffd0a.jpg)

This is a rough idea of what option 2, described in above diagram, could look like:
![allocationdetail](https://user-images.githubusercontent.com/9355962/121422156-edd03c80-c93c-11eb-8dbd-3f392dc08397.PNG)

We'll need to think about the various resource types and what attribute types they may have as well as what allocation attribute types might be available for those resource types to ensure we test all possible scenarios.  As described in the text of the first diagram, we'll need a way to define that an attribute can have a change request and if not, it is not possible to be modified in this process.  That needs to be configurable on the attribute type because admins have the ability to add new attribute types:

![allocationattributes](https://user-images.githubusercontent.com/9355962/121422715-8070db80-c93d-11eb-8332-be93ba1193c0.PNG)

PR #330, opened 2021-10-29
Allocation Change Flow

brisco17:
Resolves issue #294. This extends the Allocation system by allowing project PIs and Managers to request changes to project allocations and their attributes. 

Allocations and Allocation Attribute Types now have an "is_changeable" field that system admins can use to denote whether or not they want users to be able to request changes to those objects. If an Allocation is changeable, a user can request a date extension to that allocation. If an Allocation Attribute is changeable, users can request its value to be updated. 

To request changes, a PI should go to an Allocation's Detail Page and press the "Request Change" button in the "Allocation Information" table header which will take them to the "Change Request" page. This page takes the user's desired changes plus a justification for those changes to make an AllocationChangeRequest object (along with AllocationAttributeChangeRequest objects for every Allocation Attribute that has had a change requested, and a AllocationChangeStatusChoice object to denote the change's status).  The form also validates user requested Allocation Attribute values. 

Users can keep track of their Allocation change requests on a new "Allocation Change Requests" table on the Allocation's Detail page, which shows the date the change was requested, the change's status, any notes left by admins, and a link to a more detailed view of the request ("Allocation Change Detail" page). 

Admins can review change requests by selecting the 'Allocation Change Requests' link in the Admin tab of the navbar. The page is structured similarly to the 'Allocation Requests' page and Admins are given the option to either Approve/Deny changes on the spot, or enter their "Allocation Change Detail" pages where they can also leave notes for users. 

Users are notified by email when their changes are approved or denied, and once a change is activated the Allocation/Allocation Attributes are updated immediately.


dsajdak:
On testing found:
1. The 'Request Change' button is available for all allocations whether they're marked 'is changeable' or not.  I think this should be displayed only if the allocation is changeable or if the default to make all allocations changeable is set.
2. A PI can request a change on an expired allocation which they shouldn't be able to do.  If the allocation is locked an error is displayed and the PI can not request the change, which is what we want for expired allocations too.
3. Marked an allocation 'is changeable' and logged in as the PI on that allocation.  Clicked 'Request Change,' selected '60 days' from the date extension drop down and entered a justification.  When I clicked 'submit' I saw this error:
error
NOTE: this happens whether I change the date or modify an attribute.  I also tried doing this as an admin and get the same error.

coldfront/core/allocation/templates/allocation/allocation_change.html
```diff
@@ -0,0 +1,137 @@
...
+  you must provide a justification.
+</p>
+
+  <div class="card mb-3">
+    <div class="card-header">
+        <h3><i class="fas fa-list" aria-hidden="true"></i> Allocation Information</h3>   
+    </div>
+
+    <div class="card-body">
+      <form action="{% url 'allocation-change' allocation.pk %}" method="post">
```
aebruno, line 27: The HTML is not valid after rendering this view. Looks like we have to move the `<form>` tag up a bit. Here's a patch that will fix the imbalanced tags:

```diff
--- a/coldfront/core/allocation/templates/allocation/allocation_change.html
+++ b/coldfront/core/allocation/templates/allocation/allocation_change.html
@@ -18,13 +18,13 @@ Request Allocation Change
   you must provide a justification.
 </p>
 
+<form action="{% url 'allocation-change' allocation.pk %}" method="post">
   <div class="card mb-3">
     <div class="card-header">
         <h3><i class="fas fa-list" aria-hidden="true"></i> Allocation Information</h3>   
     </div>
 
     <div class="card-body">
-      <form action="{% url 'allocation-change' allocation.pk %}" method="post">
         {% csrf_token %}
         <div class="table-responsive">
           <table class="table table-bordered table-sm">

```

aebruno (changes requested):

dsajdak:
All issues previously reported have been resolved.  Work flow looks good and all testing passed.  One final task is to make the allocation attribute changes editable by the admin.  They should be able to approve some but not all changes.  Add a garbage can icon to the allocation attribute table.  If the admin does not approve of the change request they can either modify the value or delete it entirely.  Add a modal for 'Are you sure you want to delete?'

For the date extension, add a '0 days' or 'No extension' option to the drop down for admins to select and then no change is made to the end date of the allocation.

dsajdak (approved):
All testing passed and this new allocation change process is complete.  Great contribution!

aebruno (approved):

PR #332, opened 2021-11-03
[mokey_oidc plugin] Handle groups claim as list

knikolla:
The groups claim can be sent over as a list. The mokey_oidc plugin
is unable to deal with that, as it assumes the input for groups
is always a string.

Only call split if the claim is a string.


aebruno:
LGTM, Thanks!

Issue #336, opened 2021-11-10
Divide by zero error when Attribute that has_usage is 0


payerle:
If one has an AllocationAttribute of a type which has_usage and that attribute is set to 0, the allocation page will raise divide by zero exceptions.  The web pages should not crash.

aebruno:
Fixed in PR #337

PR #337, opened 2021-11-10
Fix for #336 - Divide-by-zero error for Attrib value=0 and has_usage

payerle:
This adds a try/except block in generate_guauge_data_from_usage() (why
the misspelling?) to catch for divide by zero errors.  If value=0, we
set the percent used to 100% (arbitrary, but better than an exception)


aebruno:
Thanks!

PR #344, opened 2021-12-22
Fix sorting arrows for allocation search

mdzik:
Was rendered as:

allocation/?order_by=end_date&direction=des&username=piospi&resource_type=**Cluster**&resource_name=1&show_all_allocations=True&

but should be this:

allocation/?order_by=end_date&direction=des&username=piospi&resource_type=**2**&resource_name=1&show_all_allocations=True&


aebruno:
I can reproduce this bug however the PR is failing for me with:

```
NameError at /allocation/

name 'Model' is not defined
```

Perhaps we can just check for the existence of the 'pk' attribute here?  For example, changing the logic to this works for me: 

```python
if isinstance(value, QuerySet):
    for ele in value:
        filter_parameters += '{}={}&'.format(key, ele.pk)
elif hasattr(value, 'pk'):
    filter_parameters += '{}={}&'.format(key, value.pk)
else:
    filter_parameters += '{}={}&'.format(key, value)
```

aebruno (changes requested):
Fix model import error and/or adjust logic to check to check for existing of pk attribute.

mdzik:
I didn't notice that we have 
import ... Model 
in our branch, sorry.


But: i still have 'a feeling' that something is not exactly right in this search (like search by date) - will do some test to confirm

mdzik:
Ok - I've found it. The problem lays in how Calendar widget handles date format. If you have different time locale in use, than date-time format needs to be compatible between Django and JS widget, and JS requires slightly different format than Django. When localization is turned on, Datepicker widget needs to have something like:

```
{% get_datepicker_format as df %}
$('#id_grant_start').datepicker({ dateFormat: "{{ df }}" });
```

where

```python
@register.simple_tag
def get_datepicker_format():
    return get_format('DATE_INPUT_FORMATS')[0].replace('%','').replace('m','mm').replace('Y','yy').replace('d','dd')

```

aebruno (approved):

mdzik:
I think that is another issue and it is not present if you don't "change language". I just left it here 'for future generations'

Issue #341, opened 2021-12-21
Port Allocation Change Request features to Allocation Requests


dsajdak:
With the allocation change request feature we modified some of the process flow.  This process should be ported back to the allocation requests for a more seamless process across the two.  What to update on allocation requests:

Allocation detail page:
- Expose button to add 'notifications' so admin can include details about allocation request
- Allow 'lock/unlock' to be set here like the 'allow change requests'

Allocation request list:
- Change 'Date Requested/Last Modified' to 'Last Modified'
- Change 'Allocation Actions' to 'Actions'
- Add 'Details' button to get to Allocation detail page
- Unlink ID number as this will now be the 'Details' button
- Grey out 'Activate' if allocation is new - this is because allocation attributes need to be added to new allocations before allocation is approved
- Remove 'Deny' button and require admin to use 'Details' to specify a reason for denying an allocation

dsajdak:
After discussion, we decided:
- Activate button on 'Allocation Requests' list will be greyed out for new allocations, forcing admin to view the allocation details
- We will not require allocation attributes but admin will get a pop-up asking if they're sure they want to activate an allocation with no attributes
- Also will not require a reason for a denial but admin will get a pop-up asking if they want to provide a reason

aebruno:
Closed in #346

PR #346, opened 2021-12-29
Ported Allocation Change Request features to Allocation Requests

brisco17:
Resolves issue #341. Added a number of features to the Allocation Request workflow that were introduced in the Allocation Change Request workflow to create a more cohesive experience. Updates include: 

Allocation detail page:

    Expose button to add 'notifications' so admin can include details about allocation request
    Allow 'lock/unlock' to be set here like the 'allow change requests'

Allocation request list:

    Change 'Date Requested/Last Modified' to 'Last Modified'
    Change 'Allocation Actions' to 'Actions'
    Add 'Details' button to get to Allocation detail page
    Unlink ID number as this will now be the 'Details' button
    Grey out 'Activate' if allocation is new - this is because allocation attributes need to be added to new allocations before allocation is approved
    Remove 'Deny' button and require admin to use 'Details' to specify a reason for denying an allocation

### DIFF
--- a/coldfront/core/allocation/forms.py
+++ b/coldfront/core/allocation/forms.py
@@ -67,6 +67,7 @@ class AllocationUpdateForm(forms.Form):
     description = forms.CharField(max_length=512,
                                   label='Description',
                                   required=False)
+    is_locked = forms.BooleanField(required=False)
     is_changeable = forms.BooleanField(required=False)
 
     def clean(self):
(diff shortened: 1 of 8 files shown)


dsajdak (changes requested):
When a new allocation has a change made and then the update button is clicked, start and end dates are added.  We would only want those added in the event the allocation status is 'active' (i.e. when the 'activate' button is clicked).  All other modifications look great!

dsajdak (approved):
LGTM

PR #347, opened 2022-01-05
Forward filter parameters between Active-Archived projects search pages

mdzik:
it seems more user-friendly if those buttons preserve search parameters


dsajdak (approved):
This is a great improvement.  Thanks @mdzik

PR #348, opened 2022-01-05
Check permissions on notification updates

mdzik:
In current state there is no login/permission check for notification update - you could disable it for all if you know how


aebruno:
Great find. Thanks for the fix.

PR #349, opened 2022-01-07
Update deploy.md

ultramathman:
I was having problems getting scheduled tasks to run after following the Deploy documentation. Discovered that Django Q requires the qcluster to be running to run scheduled tasks. Added documentation for Django Q service configuration required to make this work.


aebruno:
@ultramathman This is great! Thank you.

PR #351, opened 2022-01-12
DIRS is not an attribute but a key

lebonez:
change .DIRS to a key since this failed as below:
TEMPLATES[0].DIRS.insert(0, SITE_TEMPLATES)
AttributeError: 'dict' object has no attribute 'DIRS'


coldfront/config/plugins/ldap_user_search.py
```diff
@@ -14,4 +14,6 @@
 LDAP_USER_SEARCH_BASE = ENV.str('LDAP_USER_SEARCH_BASE')
 LDAP_USER_SEARCH_BIND_DN = ENV.str('LDAP_USER_SEARCH_BIND_DN')
 LDAP_USER_SEARCH_BIND_PASSWORD = ENV.str('LDAP_USER_SEARCH_BIND_PASSWORD')
+LDAP_USER_SEARCH_CONNECT_TIMEOUT = ENV.int('LDAP_USER_SEARCH_CONNECT_TIMEOUT', default=1)
```
aebruno, line 17: Looks like the default connect_timeout in ldap3 is set to None which ends up using a default value of 2.5. See [here](https://github.com/cannatag/ldap3/blob/dev/ldap3/utils/config.py#L82). Perhaps we should use a similar default, so either set to 2.5 or even None?

coldfront/plugins/ldap_user_search/utils.py
```diff
@@ -17,8 +17,10 @@ def __init__(self, user_search_string, search_by):
         self.LDAP_USER_SEARCH_BASE = import_from_settings('LDAP_USER_SEARCH_BASE')
         self.LDAP_BIND_DN = import_from_settings('LDAP_USER_SEARCH_BIND_DN', None)
         self.LDAP_BIND_PASSWORD = import_from_settings('LDAP_USER_SEARCH_BIND_PASSWORD', None)
+        self.LDAP_CONNECT_TIMEOUT = import_from_settings('LDAP_USER_SEARCH_CONNECT_TIMEOUT', 1)
```
aebruno, line 20: Same here, see previous comment on default value.

aebruno (changes requested):
@lebonez Thanks for the fixes! Just spotted a very minor change with the defaults to better match those from ldap3.

coldfront/config/plugins/ldap_user_search.py
lebonez: Agreed 2.5 is probably more reasonable.

aebruno (approved):

aebruno:
Thanks!

Issue #355, opened 2022-01-14
Allocations:  Change request does not require justification


dsajdak:
A PI should have to provide a justification for the change request they're submitting.  Also, users (non-managers) can see the date drop down in a pending allocation change request.  They can toggle the dates but can't save it or otherwise change it.  This was previously fixed so that PIs could not access that.  They see the date extension they requested and it is greyed out.

aebruno:
Closed in #362

PR #362, opened 2022-01-18
Allocation Change Flow Bug Fixes

brisco17:
Resolves issue #355. 

- Justifications are now required again for any allocation change request that is submitted.
- Users (non-managers) can no longer see the date drop down in a pending allocation change request.


dsajdak (approved):
This all looks good!

PR #364, opened 2022-01-20
Added signal dispatch for allocation change requests

brisco17:
Signals are now dispatched for

- allocation change request approval


dsajdak (approved):
My basic testing is not finding any issues with this.

PR #365, opened 2022-01-26
allocation list view if project is New or Active

lebonez:
Show allocation in list view if project is New or Active not just Active.


dsajdak:
@lebonez I'm not sure I'm following what you're suggesting.  Under the Project menu and Allocations list an admin can click on Filter and then 'Show all allocations' to see everything or they can use the Admin menu and click on "All Allocations" which redirects to this Allocations list and the 'Show all allocations' box is already checked.  Is this what you mean?

dsajdak:
@lebonez thanks for the contribution.  We didn't realize there was a filter issue on the allocations list page for users.

PR #367, opened 2022-01-31
Fix IntegrityError when publications have same UniqeID and Project

mdzik:
but different other fields, like source or for example abbreviation changes in names etc. Instead of  meaningful info, IntegrityError is raised while Django tries to insert a new record.

get_or_create uses all fields in select, defaults are used instead for new records.


aebruno:
Related issue #369

dsajdak (approved):
This looks good and resolves an error reported by a CCR user recently.  Thank you @mdzik!

Issue #292, opened 2021-05-03
Allocations:  Allow allocation users to see allocations of all statuses


dsajdak:
Currently users on an allocation can't see it on their project if it's in the status of 'payment declined,' 'payment pending,' 'pending,' 'unpaid,' 'revoked,' 'approved,' or 'denied'.  If they navigate directly to the allocation detail URL they see the error "you are not authorized to view this page"

PR #376, opened 2022-02-19
Project Detail Page: Users can see all their allocations (regardless of status)

brisco17:
Resolves issue #292

- On an active Project's detail page users are now able to view all the allocations they have access to, regardless of their current status. 
- Also fixed a minor UI bug on Allocation Detail Pages where staff users used to get active checkbox fields for the "Lock/Unlock Allocation" and "Allow Change Requests" fields. Those are now disabled for them.


dsajdak (approved):
This is good to go

Issue #283, opened 2021-03-29
Publication:  item with many authors fails to get added


dsajdak:
Trying to add a publication with many authors maxes out the field.  Perhaps pull the lead author and concat with "et al" when we encounter something like this?

![coldfront_error](https://user-images.githubusercontent.com/9355962/112893931-45cf0500-90a9-11eb-9d41-f0ea48d9ecf3.PNG)

Example for testing:  10.1103/PhysRevD.102.032007

Issue #247, opened 2020-11-17
Publications:  Anything with special characters in the title fail to be added to project


dsajdak:
Publications with special characters in the title fail to be added to the project.  Examples of these characters include scientific formulas.

User gets error:  We're having a bit of system trouble at the moment. Please check back soon!

From logs:  django.db.utils.OperationalError: (1267, "Illegal mix of collations (latin1_swedish_ci,IMPLICIT) and (utf8_general_ci,COERCIBLE) for operation '='")

dsajdak:
First step would be to improve this messaging.  If a publication can't be added, the error should say that, rather than just "system trouble."  Something like "This publication could not be added.  If there are special characters or scientific formulas in the title, please add the publication manually."

aebruno:
This is because currently we're not setting the default charset and in mariadb/mysql 5.5 it's defaulting to latin1. We really need to change this to utf8mb4 for mysql/mariadb. The short term fix is to run this directly on the db: ```ALTER TABLE publication_publication CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci;```

PR #377, opened 2022-03-01
Bug Fix: Publications can be added to projects no matter how many authors they have

brisco17:
Resolves issue #283
* If a publication is searched for by its ID in order to be added to a project, it will now be successfully added no matter how long its author field is.
* Publications like this will have its author field truncated down to 1024 characters.

### DIFF
--- a/coldfront/core/publication/forms.py
+++ b/coldfront/core/publication/forms.py
@@ -22,7 +22,7 @@ class PublicationSearchForm(forms.Form):
 
 class PublicationResultForm(forms.Form):
     title = forms.CharField(max_length=1024, disabled=True)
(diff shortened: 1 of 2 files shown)


dsajdak (changes requested):
These changes fixed the error for adding a publication with an author list longer than 1024 characters.  However, once the publication is added, it can't be deleted.  

puberror

aebruno:
This may be related to #247

dsajdak (approved):

dsajdak:
This error was related to the mariaDB/mysql utf8 support.  Other testing shows these changes are good so we're going to approve it and merge.  We'll be recommending Postgres for better encoding support, specifically for scientific and math characters frequently used in publication titles.

Issue #222, opened 2020-06-30
grant export: download only those found under search


dsajdak:
We can search for grants in the grant report but the 'export' button downloads everything.  would be nice to be able to export only those found in the search

dsajdak:
![](https://github.trello.services/images/mini-trello-icon.png) [Grant export](https://trello.com/c/ZVCG6CjW/1-grant-export)

dsajdak:
In addition to this issue, publications with long titles break the export:
![pubexport](https://user-images.githubusercontent.com/9355962/109556439-d44d6800-7aa4-11eb-82ff-c12f9e664efb.PNG)

PR #382, opened 2022-03-15
Export Specific Grants

brisco17:
Resolves issue #222

- Each grant on the 'Grant Resource' page now has a checkbox that can be used to export specific resources.
- Resources can now be duplicated in the Django Admin portal through the use of a 'Save as new' button.


dsajdak (approved):
Looks good to me!

Issue #386, opened 2022-04-04
Allocations:  Users are added to inactive allocations


dsajdak:
A bug was introduced with recent updates.  New users added to a project can be added to allocations in the 'inactive (renewed)' state.  The PI/manager sees allocations in the 'active' and 'inactive(renewed)' states as options to add the user to.

aebruno:
Resolved in #390

PR #390, opened 2022-04-21
Bug Fix: Users can no longer be added to inactive allocations when being added to a Project

brisco17:
Resolves issue #386 

- When a PI tries to add a user to a Project, they now can only add the user to Project allocations of the following statuses: Active, New, Renewal Requested, Payment Pending, Payment Requested, Paid


dsajdak (approved):
This looks good to me

PR #392, opened 2022-04-26
SLURM_NOOP should be a bool

lebonez:
SLURM_NOOP should be a bool. "if self.noop" resolves as true even if self.noop is "False" because it is a non-empty string. Noticed this because slurm_check was responding as NOOP even when using --sync.


aebruno:
Thank you!

PR #399, opened 2022-05-25
Project attributes.

RyanChang0369:
Added project attributes that behave like allocation attributes. The attribute options are just placeholders for now, so they need to be replaced.


coldfront/core/project/templates/project/project_attribute_delete.html
```diff
@@ -0,0 +1,73 @@
...
+                  <td>{{ form.name.value }}</td>
+                  <td>{{ form.value.value }}</td>
+                </tr>
+              {% endfor %}
+            </tbody>
+          </table>
+        </div>
+        {{ formset.management_form }}
+        <div>
+          <button type="submit" class="btn btn-danger"><i class="fas fa-minus" aria-hidden="true"></i> Delete Selected Attributes from
```
aestoltm, line 44: Receiving a 'str' object has no attribute 'get' error when deleting project attributes. They delete on the backend and show that in the frontend but need to back out of the error page to see the changes.

coldfront/core/project/templates/project/project_detail.html
```diff
@@ -151,6 +151,45 @@ <h3 class="d-inline" id="users"><i class="fas fa-users" aria-hidden="true"></i>
 </div>
 <!-- End Project Users -->
 
+<!-- Start Project Attributes -->
+<div class="card mb-3">
+  <div class="card-header">
+    <h3 class="d-inline"><i class="fas fa-info-circle" aria-hidden="true"></i> Attributes</h3> <span class="badge badge-secondary">{{attributes.count}}</span>
+    <div class="float-right">
+      {% if project.status.name != 'Archived' and is_allowed_to_update_project %}
```
aestoltm, line 159: Managers do not have permissions to create project attributes. Only admin are able to at the moment.

aestoltm (changes requested):
@RyanChang0369 We appreciate your contribution to Coldfront! We have some changes to request before we can merge in your pull request. 

**The following changes would be appreciated:**
1. Adding project attribute choices through a migration instead of loading them with the 'add_default_project_choices' command (this will help existing users have the attribute choices without running the command). The command currently is meant for fresh installs and will wipe any projects for existing users.
2. In the administration page, including a section for project attribute choices so they can be set as is_private, is_changable, etc and so attributes can be created/deleted through the admin page.
3. In the administration page, linking project attributes to a project similar to how you can view allocation attributes on an allocation. Project attributes should be able to be created/deleted and edited through this page.
4. Adjusting the look for 'is_private' project attributes to have a grey cell color and lock symbol for the table (similar to how allocation attributes are handled)
5. Deleting project attributes responses with a **'str' object has no attribute 'get'** error, the project attributes do delete but you need to back out of an error page to see the changes.
6. Allowing managers to add/delete project attributes. Managers do not have permission to add/delete project attributes, only admins are able to do this at the moment.
7. Updating the informational hover to mention project attributes for the 'Role' column header in the 'Project Detail' page for users on a project.

Thank you again for your contribution, we are looking forward to be able to include your changes into Coldfront!

dsajdak:
@RyanChang0369 and @aestoltm I just want to clarify item no. 6 above.  This would be similar to the functionality that we have for the ability to request the change of allocation attributes (see https://coldfront.readthedocs.io/en/latest/manual/allocations/change_requests/).  An admin can set an allocation attribute type to "is changeable" and then the project manager is allowed to request a change on that attribute.  This functionality may be more work than you want to get into for this PR.  We can tackle that after merging in these other changes.  Thanks again for your contributions!

PR #401, opened 2022-06-01
Resolved #387: Fixed bug that would only update allocation statuses o…

### LINKED ISSUE #387: Allocations:  Lock allocations in expired, denied, & inactive(renewed) status
When an allocation is set in expired, denied, or inactive (renewed) states the allocation should be locked to prevent a PI from adding new users to it.  The allocation can be unlocked in the ColdFront administration UI by an admin, if necessary.

aestoltm:
…f 'active' to 'expired', other statuses would stay as their current status after expiration date was reached.

NOTE: Need redis running and possibly MailHog to have the scheduler work

Tested fix through the following: 
1. Created many allocations of different statuses (also tested some allocations that were locked/unlocked and some that allowed change requests)
2. Put all created allocations start and end dates so that they should be expired
3. Through ColdFront Administration page, forced the DJANGO Q 'Scheduled tasks' to perform a status update.
4. Force status update by changing 'Date' to 'Now' and 'Time' to 'Now', then click 'SAVE'
5. Status updates should occur shortly after. Then check allocations for the expected status update.

Tested bug through the following:
1. Tested 'active' status to 'expired' 
2. Tested bug of other statuses to 'expired' (found they did not switch their status as described in bug report)
3. Follow steps 3-5 above to force scheduler to update statuses

### DIFF
--- a/coldfront/core/allocation/tasks.py
+++ b/coldfront/core/allocation/tasks.py
@@ -31,7 +31,7 @@ def update_statuses():
     expired_status_choice = AllocationStatusChoice.objects.get(
         name='Expired')
     allocations_to_expire = Allocation.objects.filter(
-        status__name='Active', end_date__lt=datetime.datetime.now().date())
+        end_date__lt=datetime.datetime.now().date())
     for sub_obj in allocations_to_expire:
         sub_obj.status = expired_status_choice
         sub_obj.save()


dsajdak:
This is going to break the Invoice report but it needs to be done.  On the roadmap is a rework of how we handle "invoices" or payments.  If an invoice isn't paid on time, the allocation should be set to expired and access to the resource should be blocked.

aebruno:
One side effect of this change is that all allocations with an end_date in the past will now be changed to expired every time this job is run. Here's a few cases where this may not be desired: 

- Expired allocations that already have their status set to expired will get updated unnecessarily. 
- If an admin wanted to explicitly set the status of an expired allocation to something other than expired, this will get overwritten next time this job runs. With this change there will be no way to override this. 
- Sites that have added custom statuses may have different business logic that conflicts with this change

Perhaps a better approach would be to add other statuses here in the filter instead of removing the "Active". So maybe we come up with the list of statuses an allocation must be along with an end_date in the past that will trigger this change of status to "expired". That will make this job more predictable and allow other sites to modify the behavior if desired. Thoughts?

aebruno (approved):
LGTM

dsajdak (approved):
Tested functionality and LGTM

PR #402, opened 2022-06-02
Fixed allocation request 'approved' status bug

### LINKED ISSUE #379: Allocations:  approved status
Currently the allocation status of "approved" has no use other than for business process.  One use case would be for centers that have allocation review committees.  The committee would review and approve the allocation request and then the admin would set attributes and activate the allocation.  Currently when we change the status to 'approved' it disappears off the Allocations Request list.  Let's change the allocations request list to also display the 'approved'  status as well.

aestoltm:
Resolved #379 : Fixed bug that would cause allocations that had status of 'New', 'Renewal Requested', or 'Paid' would disappear from the allocation request page when changed to a status of 'Approved'. 

Tested bug through the following:
1. Created allocation requests that had statuses of 'New', 'Renewal Requested', and 'Paid' and changed their status to 'Approved'.
2. Confirmed that these allocation requests would not appear on the allocation request page after status change.

Tested fix through the following:
1. Followed the steps for the bug testing and confirmed that these allocations remained on the allocation request page after fix was applied.

Source of bug:  AllocationRequestListView class of the views file for Allocation would only list allocations with status names of 'New', 'Renewal Requested', and 'Paid'. Added 'Approved' into the status names included.

### DIFF
--- a/coldfront/core/allocation/views.py
+++ b/coldfront/core/allocation/views.py
@@ -1095,7 +1095,7 @@ class AllocationRequestListView(LoginRequiredMixin, UserPassesTestMixin, Templat
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         allocation_list = Allocation.objects.filter(
-            status__name__in=['New', 'Renewal Requested', 'Paid', ])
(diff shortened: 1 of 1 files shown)


dsajdak (approved):
Testing passed.  LGTM

PR #403, opened 2022-06-06
Add 'notes' button on project detail page

### LINKED ISSUE #194: Projects: admin messages not displayed, no easy way to add them in UI
When updating an allocation, admins should be able to add a new message to the users.  There is no way to do this in the UI, only in the Django backend.  Even after adding an admin message on an allocation in the backend, it is not displaying in the allocation details in the UI.

aestoltm:
Resolved #194 : Added feature to allow notes to be added on project detail page through a button. This feature is essentially identical to how allocation notes work on the allocation detail page.

Tested feature through the following:
1. Created project notes on an admin account (both private and public project notes). 
2. Ensured that the project note created was displayed the 'note', 'author' and 'date' (for both private and public).
3. Ensured that non-admin accounts could only see public project notes. Also that non-admin accounts could not create project notes.


aebruno:
Once we flesh out #407 this will need to get re-worked a bit. Just wanted to add a link so we remember.

dsajdak (approved):
Testing passed.  This looks good to me.  It should be pointed out that this functionality allows the admin to add a public or private note to a project.  If the admin adds a note on the project in Django, this is a private note and therefore not visible to project users.  There is no way to make the note public in the Django admin.

Issue #393, opened 2022-04-28
Fix alphabetical sorting


dsajdak:
Fix alphabetical sorting of resources in allocation request drop down list and in the django admin list of resources so that it is case insensitive.  Also fix needed in allocation and resource attribute listings.

PR #406, opened 2022-06-08
Issue #393 resolved (lists are alphabetically sorted case insensitive)

rg663:
In the allocation request drop down list, django admin list of resources, and allocation and resource attribute listings, values are now listed alphabetically case insensitive.

### DIFF
--- a/coldfront/core/resource/views.py
+++ b/coldfront/core/resource/views.py
@@ -3,6 +3,7 @@ from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.db.models import Q
+from django.db.models.functions import Lower
 from django.forms import formset_factory
 from django.http import HttpResponseRedirect, JsonResponse
 from django.shortcuts import get_object_or_404, render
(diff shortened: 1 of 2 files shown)


dsajdak (changes requested):
In testing, I see that the list of resources in the drop down menu when requesting a new allocation is now sorted alphabetically regardless of case.  I also see this in the 'All resources" list in the UI.  However, allocation and resource attributes are not sorted in a case insensitive manner and the resources listed in the Django admin are also not sorted this way.  In this PR I'm only seeing changes to the allocation/forms.py and resource/views.py files.  @rg663 are there changes that you meant to include in the PR that aren't here?

coldfront/core/resource/views.py
```diff
@@ -197,7 +200,16 @@ def get_queryset(self):
 
         if resource_search_form.is_valid():
             data = resource_search_form.cleaned_data
-            resources = Resource.objects.all().order_by(order_by)
+            if order_by == "name":
```
aebruno, line 203: I wonder if we can refactor the logic a bit here? It might be easier to read if we organized it like this:

```
        # These are used elsewhere below so lets initialize them here. We can
        # use the default value of the get() method to set defaults
        order_by = self.request.GET.get('order_by', 'id')
        direction = self.request.GET.get('direction', 'des')
        if order_by != "name":
            if direction == 'des':
                direction = '-'
            order_by = direction + order_by
```

Would something like that work?

aebruno (changes requested):

PR #422, opened 2022-06-30
Arguments in signal definitions are obsolete

mdzik:
https://stackoverflow.com/questions/70364068/django-cms-unexpected-keyword-argument-providing-args


dsajdak (approved):
Looks good to me!

PR #423, opened 2022-06-30
Add GPUh and Storage metrics support for XDMOD import

mdzik:
This will import GPUh and GB to respective metrics from XDMOD instance. It was tested against our XDMOD, but please be careful. 

Additional argument for xdmod_usage will import metrics for expired allocations - useful for new/migrated setups.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ coldfront.db
 .vscode
 db.json
 .env
+.devcontainer/*
+.gitignore
+.bin/*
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -19,3 +19,6 @@ coldfront.db
 .vscode
 db.json
 .env
+.devcontainer/*
+.gitignore
```
aebruno, line 23: I don't think we want to gitignore the gitignore file here :) Can you remove?

aebruno (changes requested):

.gitignore
mdzik: No we don't. :)

aebruno (approved):
LGTM

PR #424, opened 2022-07-01
Notifications for warranty and service expirations on resources

### LINKED ISSUE #371: Notifications for resources
- Add the option to configure CF to email PIs when their equipment is hitting either of these dates.  By default this should be off.
- Allow the configuration of what X is in CF like we do with allocation renewals
- Display notification on resource detail page when resource has a 'WarrantyExpirationDate' in less than X days/months 
- Display notification on resource detail page when resource has a 'ServiceEndDate' in less than X days/months 
- Display notification on resource detail page when children have a 'WarrantyExpirationDate' in less than X days/months so PIs know they have equipment going off warranty soon.
- Display notification on resource detail page when children have a 'ServiceEndDate' in less than X days/months so PIs know they have equipment being decommissioned soon.

aestoltm:
Resolved #371 : Provided email and webpage message notifications for resource warranty and service attribute expirations based on a user defined set of days similar to how 'EMAIL_ALLOCATION_EXPIRING_NOTIFICATION_DAYS' works. Also, users have the option to disable email notifications with 'EMAIL_RESOURCE_NOTIFICATIONS_ENABLED' (defaulted to off). 

Tested:
1. Set end dates for warranty and service attributes for resources with and without children resources.
2. Ensure that dates that are less than or equal to the set dates in 'EMAIL_RESOURCE_EXPIRING_NOTIFICATION_DAYS' will cause a message to appear on the webpage when viewing the resource on the resource detail page.
3. Ensure that emails are sent (if enabled) when end dates are equal to those set in 'EMAIL_RESOURCE_EXPIRING_NOTIFICATION_DAYS'.
4. Ensure webpages work for resources that do not have service and warranty end dates set or their end dates are not soon enough to generate notifications.


dsajdak:
Upon further discussion we have decided to revamp the logic on determining who should be notified when a resource warranty or service end date is approaching.  We will add a resource attribute called owner_username.  That username will be looked up in ColdFront to find their email and then send the email.  This completely eliminates anything related to projects, allocations, or PIs, at least for email notifications.  We will still display the banner on the project and allocation detail pages for any resources or child resources that have the warranty and/or service date approaching.  This way the information is accessible to anyone who is on an allocation for these affected resources.

PR #440, opened 2022-07-26
update README with location of standalone starfish plugin repo

claire-peters:
Hey all, I made a standalone repo for the starfish plugin! FASRC-specific configurations have been removed and the repo is tested for integration with the UBCCR repo master branch.


aebruno:
@claire-peters Thank you!

PR #444, opened 2022-08-05
Switch to LDAP lookup for nsaccountlock in freeipa.

### LINKED ISSUE #416: FreeIPA plugin: not recognizing usernames greater than 11 characters
When running the freeipa plugin to sync groups, a username that has 12 or more characters in it is erroring out.

`ipa: WARNING: User dsajdaktest should be added to freeipa group: academic`
`dsajdaktest     academic                Enabled Active`
`ipa: WARNING: User jbednasztest not found in FreeIPA`

However, using the freeipa command (ipa group-add-member) to add the user to the group works.  

I would say this is a high priority to fix as it's easy to miss this error if running in a cron.

Test Tasks:
1. Add a new user to FreeIPA with a username of 12 characters or more
2. Add user to an allocation that has the freeipa_group attribute
3. Run the freeipa sync for that user to see if the account is added to the group: 
`coldfront freeipa_check -x -s -u username`

aebruno:
Fixes #416

### DIFF
--- a/coldfront/plugins/freeipa/management/commands/freeipa_check.py
+++ b/coldfront/plugins/freeipa/management/commands/freeipa_check.py
@@ -8,6 +8,7 @@ from django.core.management.base import BaseCommand, CommandError
 from ipalib import api
 from ipalib.errors import NotFound
 
+from coldfront.plugins.freeipa.search import LDAPUserSearch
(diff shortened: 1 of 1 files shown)


dsajdak (approved):
Tested by creating a new account with username 'superlongusername' and the FreeIPA syncing worked correctly.

PR #445, opened 2022-08-05
Lists sorted case insensitive throughout front end

### LINKED ISSUE #393: Fix alphabetical sorting

rg663:
Resolves #393 — lists are now sorted CI (alphabetically regardless of case) throughout ColdFront's front end. No changes were made to models or any part of the admin site, so no migrations need to be made.

Instances that were fixed:

1. When adding users to projects, previously, the allocation list for the project was sorted out of order. Now, it is sorted case insensitive.
2. When requesting a resource allocation, the resource list was sorted alphabetically, but not CI. Now, it is sorted case insensitive.
3. When adding a grant to a project, the grant funding agencies were not sorted alphabetically. Now, they are sorted alphabetically (CI doesn't apply since all agencies are in title case).
4. When filtering allocations on the list page, the resource names and allocation attribute names were sorted alphabetically case sensitive. Now, they are sorted case insensitive.
5. When adding an allocation attribute, the attribute names were not sorted CI, but now they are sorted case insensitive.
6. When sorting by resource name on the resource list page, resources are now sorted case insensitive as opposed to just alphabetically.
7. When adding a resource attribute to a resource, previously, the list of attributes was not sorted case insensitive. Now, it is sorted case insensitive.


coldfront/core/allocation/views.py
```diff
@@ -923,8 +921,9 @@ def post(self, request, *args, **kwargs):
 
 class AllocationAttributeCreateView(LoginRequiredMixin, UserPassesTestMixin, CreateView):
     model = AllocationAttribute
+    form_class = AllocationAttributeCreateForm
     # fields = ['allocation_attribute_type', 'value', 'is_private', ]
-    fields = '__all__'
+    # fields = '__all__'
```
aebruno, line 926: Let's remove this comment.

coldfront/core/user/forms.py
```diff
@@ -1,4 +1,6 @@
 from django import forms
+from django.db.models.functions import Lower
```
aebruno, line 2: Do we use this anywhere in this file? If not, we can remove the import

aebruno (changes requested):

dsajdak (approved):
All of these changes look great.  The experience for user and admin will be much nicer now.  There are a couple of other things we can address to make the lists more consistent across the board but I will open new issues for those.  They're really minor and don't directly impact these updates.

aebruno (approved):

PR #450, opened 2022-08-18
Project Attributes Feature Request

aestoltm:
Draft PR that is built off PR (#399) created by @RyanChang0369.

Issues that need to be addressed with this PR are commented in (#399). Additional changes are possible as work continues on this feature.


coldfront/core/project/management/commands/add_default_project_choices.py
```diff
@@ -27,3 +29,15 @@ def handle(self, *args, **options):
         ProjectUserStatusChoice.objects.all().delete()
         for choice in ['Active', 'Pending - Add', 'Pending - Remove', 'Denied', 'Removed', ]:
             ProjectUserStatusChoice.objects.get_or_create(name=choice)
+        
+        for attribute_type in ('Date', 'Float', 'Int', 'Text', 'Yes/No'):
+            AttributeType.objects.get_or_create(name=attribute_type)
+
+        ProjectAttributeType.objects.all().delete()
```
aebruno, line 36: We shouldn't delete all here. This will remove all projects, attributes etc. Instead, we should only add any NEW defaults related to this PR.

aebruno (changes requested):

coldfront/core/project/management/commands/add_default_project_choices.py
dsajdak: Refer to #1 here: https://github.com/ubccr/coldfront/pull/399

aebruno:
Replaced by #466

PR #454, opened 2022-09-16
Attribute type validator

### LINKED ISSUE #451: Attribute type validation error
If an attribute is added using the incorrect type (text in an INT field), an error is thrown.  This is coming up as we're incorporating project attributes too.  Will be working to build a validation tool.

aestoltm:
Fixes #451 : Provided a validation tool for attribute types. The validation tool can be expanded to accommodate more data types as necessary. Instead of having to write custom validation of attribute types, now you can create an AttributeValidator class and use the provided methods for validating attribute types. This structure will help adhere to DRY principle of software development.

Tested implementation by:
- Testing incorrect and correct values for each attribute type when creating and updating attributes


aebruno:
Some of this code exists in #466 which will be merged first.

PR #455, opened 2022-09-21
Description Bug when Approving an Allocation Request

### LINKED ISSUE #433: Allocation:  Approve button deletes description text
When an admin adds text into the description box on the allocation detail page it is wiped out if they click on the 'approve' or the 'deny' button.  The 'update' button works as intended.  

Task Tests:

1. Sign into ColdFront as the admin user
2. Create a new allocation request
3. Navigate to the allocation requests list and click on the 'Details' button for the pending request
4. Enter a description and click the 'Approve' or 'Deny' button
5. Verify the description text remains and is unchanged.

### LINKED ISSUE #470: Allocation:  Approve button resets date selection
When the approve button is clicked, the status is changed to 'active,' the start date is set to today and the end date is set to 1 year from today (or whatever ALLOCATION_DEFAULT_ALLOCATION_LENGTH is set to).  The admin should be able to set a different start and/or end date but can only do this if they do not use the 'Approve' button.  When an admin selects a start date and/or end date on the allocation detail page, it is wiped out if they click on the 'approve' button. The 'update' button will save the changes properly but if the approve button is then used to change the status, it overwrites the selected dates. 

Task Tests:

- Sign into ColdFront as the admin user
- Create a new allocation request
- Navigate to the allocation requests list and click on the 'Details' button for the pending request
- Select a start and/or end data and click the 'Approve' button
- Verify the date(s) selected remains and is unchanged and that the status is changed to 'Active.'
- If a start date is selected but no end date, the process should automatically set the end date to whatever ALLOCATION_DEFAULT_ALLOCATION_LENGTH is set to in coldfront.env (default is 365 days)

aestoltm:
Fixes #433 : Fixed description not displaying when setting value before clicking 'Approve' for an allocation request. Also Fixes #470 

Tested fix:
1. Approved allocation without any information to ensure that start_date and end_date are automatically applied
2. Approved allocation without any information to ensure that start_date and end_date are automatically applied along with description text
3. Approved allocation with custom start_date and end_date along with description text
4. Approved allocation with custom start_date or end_date along with description text (should have start_date and end_date automatically applied in this case)


dsajdak (changes requested):
This solves the issue of the text being wiped out when hitting the approve button but the bug is also there when clicking the 'deny' button.  Interestingly, your testing brings up another issue.  There is no way to set a custom start and/or end date when clicking the 'approve' button.  If dates are entered in either or both of those boxes, they're wiped out and replaced by start date of today and end date of one year from today.  We should have logic built in that sets the dates like that only if dates are not already selected.  If there's a start date entered and no end date, the end date should be set one year from the start date (or whatever is set in the coldfront.env config file).  This should be a separate git issue though so I will create that.  However, the issue with the text in the description box being wiped out when clicking the deny button should be fixed as part of this PR, please.

coldfront/core/allocation/templates/allocation/allocation_detail.html
```diff
@@ -164,7 +164,7 @@ <h3><i class="fas fa-list" aria-hidden="true"></i> Allocation Information</h3>
       <div class="float-right">
         {% if allocation.status.name == 'New' or allocation.status.name == 'Renewal Requested' %}
           <button name="approved" class="btn btn-success mr-1 confirm-activate">Approve</button>
-          <a href="{% url 'allocation-deny-request' allocation.pk %}" class="btn btn-danger mr-1 confirm-deny">Deny</a>
+          <button name="denied" class="btn btn-success mr-1 confirm-deny">Deny</button>
```
dsajdak, line 167: changing this to btn-success changed the color of the button from red to green

dsajdak (changes requested):
Just one minor change requested on this - button got changed to green, I think accidentally.  The functionality works as desired though.

dsajdak (approved):
LGTM!

PR #460, opened 2022-09-27
Added warning to initial_setup script

### LINKED ISSUE #435: Installation:  Add warning to initial setup script
As part of the installation, we run `coldfront initial_setup` which sets up defaults like attribute types and other example info.  We do not recommend using this for existing installations and want any changes to ColdFront code to be added to the migrations functionality.  This will allow existing users who upgrade ColdFront to get the necessary changes but not affect their existing data.  We do want to add a warning message to this initial_setup just in case someone runs this without realizing it will delete existing data.  Some like:

"WARNING:  Running this command initializes the ColdFront database.  This should only be run one time.  Running this will delete anything currently in the ColdFront database.  Do you want to proceed?  Yes/No"

aestoltm:
Fixes #435 : Added (Yes/No) choice and warning when running `coldfront initial_setup` script.


coldfront/core/utils/management/commands/initial_setup.py
```diff
@@ -8,13 +8,31 @@
 
 
 class Command(BaseCommand):
+    help = 'Run setup script to initialize the Coldfront database'
+
+    def add_arguments(self, parser):
+         parser.add_argument("-fo", "--force_overwrite", help="Force intial_setup script to run with no warning.")
```
aebruno, line 14: Let's make this a boolean arg by adding `action="store_true"`. Also, let's rename the short flag to just `-f`.

coldfront/core/utils/management/commands/initial_setup.py
```diff
@@ -8,13 +8,31 @@
...
-        call_command('add_default_project_choices')
-        call_command('add_resource_defaults')
-        call_command('add_allocation_defaults')
-        call_command('add_default_publication_sources')
-        call_command('add_scheduled_tasks')
+            if options['force_overwrite']:
+                run_setup()
+
+            else:
+                print("""WARNING: Running this command initializes the ColdFront database. This should only be run one time. Running this will delete anything currently in the ColdFront database.""")
```
aebruno, line 21: Can we tweak the verbiage here to be:

> WARNING: Running this command initializes the ColdFront database and may modify/delete data in your existing ColdFront database. This command is typically only run once.

aebruno (changes requested):
Looks good. Just had a few minor tweaks.

aebruno (approved):

PR #463, opened 2022-09-27
Allocation Detail Page User Sorting

### LINKED ISSUE #409: Allocations: Add User Sorting
In addition to sorting by username (as completed in #404 we should also be able to sort by users' first and last names.  The name field may need to be broken into two to allow this to work.  This may be related to work required for issue #408

aestoltm:
Fixes #409 : Added sorting to the allocation user section in the allocation detail page.


dsajdak (approved):
This works well!

Issue #474, opened 2022-12-12
The Denied status of an allocation is the only status that sends a signal


knikolla:
There is an `allocation_disable` signal, but it is only sent for the `Denied` status. https://github.com/ubccr/coldfront/blob/68ba4b62ea5e47cf2295dede1f3df13868ec4788/coldfront/core/allocation/views.py#L251

It should at least also be sent for the `Revoked` status, and perhaps some other relevant status.

The motivation is that it does seem a bit weird to "deny" an allocation that already exists, rather than "revoking" it.

PR #476, opened 2023-01-06
Signal fixed on changing allocation statuses

rg663:
Resolves issue #474 — Now, allocations that have been given a revoked status display a similar behavior to an allocation that has been marked as denied, sending an email to the user and signaling that the allocation should be disabled. In addition, denying an allocation that already exists, which is counterintuitive behavior, displays a warning to the user alerting them of this action.


coldfront/templates/email/allocation_revoked.txt
```diff
@@ -0,0 +1,10 @@
+Dear {{center_name}} user,
+
+Your allocation request for {{resource}} has been revoked.
```
dsajdak, line 3:
```suggestion
Your allocation for {{resource}} has been revoked.
```
Let's remove "request" since we're only revoking an existing, likely active, allocation, not a new allocation or renewal request.

coldfront/templates/email/allocation_revoked.txt
```diff
@@ -0,0 +1,10 @@
...
+
+Please login to view a message from the system administrators: {{allocation_url}}
+If you are a student or collaborator under this project, you are receiving this notice as a courtesy. If you would like
```
dsajdak, line 6: Add a new line in between these two paragraphs.  I will update the other email templates this way too

coldfront/core/allocation/views.py
```diff
@@ -257,7 +258,11 @@ def post(self, request, *args, **kwargs):
                     allocation_remove_user.send(
                         sender=self.__class__, allocation_user_pk=allocation_user.pk)
 
-                send_allocation_customer_email(allocation_obj, 'Allocation Denied', 'email/allocation_denied.txt', domain_url=get_domain_url(self.request))
+                if new_status == 'Denied':
+                    messages.warning(request, "Denying an allocation that already exists. Did you mean to revoke the allocation instead?")
```
aebruno, line 262: Curious why are we issuing this message? Is this case handling the user "Denying" and allocation that is already in "Denied" status?

dsajdak: I do like this idea but it opens up a can of worms because we don't do this in other places.  I think for now, let's remove it to keep it consistent with other status work flow.

aebruno: Sounds good. @rg663 this PR looks good, if you can just remove that message we can get this merged. Thanks!

coldfront/core/utils/mail.py
```diff
@@ -102,7 +102,7 @@ def send_allocation_admin_email(allocation_obj, subject, template_name, url_path
     ctx = email_template_context()
     ctx['pi'] = pi_name
     ctx['resource'] = resource_name
-    ctx['url'] = url
+    ctx['allocation_url'] = url
```
aebruno, line 105: This function is used by lots of other templates. So you can't change the key here. Keep this as `ctx['url']` and you just need to update your new templates to use `{{url}}` instead of `{{allocation_url}}`

aebruno (changes requested):

PR #484, opened 2023-02-04
Refactor allocation approval.

### LINKED ISSUE #433: Allocation:  Approve button deletes description text

aebruno:
This commit refactors the allocation approval logic in the following ways:

1. Removes the AllocationActivateRequestView and AllocationDenyRequestView. These views are currently handled via GET requests and because they modify the database they really should be POST requests. The logic in these views is also a duplicate of what is already contained in the AllocationDetailView.

2. Modify AllocationDetailView to handle the various actions of update, approve, auto-approve and deny allocations. These are defined as follows:

   - update = user clicks the "update" button on the allocation detail page with the intent of updating the allocation data.

   - approve = user clicks the "approve" button on the allocation detail page. In addition to updating the allocation data this will set the status to "Active" and set the end date equal to now + ALLOCATION_DEFAULT_ALLOCATION_LENGTH days This also properly fixes #433.

   - auto-approve = user clicks the "approve" button from the allocation request list page. This will set the status to "Active" and set the end date equal to now + ALLOCATION_DEFAULT_ALLOCATION_LENGTH days. It will not update any other data on the allocation.

   - deny = user clicks the "deny" button on the allocation detail page. This will set the status to "Denied" and remove the start/end dates.

3. On the allocation request list page, the "approve" button now submits a POST request instead of a get request to "auto-approve" the allocation (see above logic). Also the button label has been changed to "Approve" to be consistent with the allocation detail page.


dsajdak (changes requested):
This broke part of the allocation change request approval process.  If we request a date extension on the allocation, when I go to approve that extension it only changes the date if I click on the 'approve' button from the allocation change requests list page ("auto-approve").  If I click the 'approve' button from the allocation change request detail page it does not change the date.  Other than that, everything else works well.

aebruno:
> This broke part of the allocation change request approval process.

Nothing in this PR touched the allocation change request approval process. It appears this was broken before so would rather fix that in a separate PR. Can you confirm you see this behavior in the main branch? If so, we can fix in another PR.

dsajdak (approved):

PR #488, opened 2023-02-08
Fix logic in allocation change request flow.

aebruno:
- Remove activate/deny GET request views for change requests
- Fix bug when setting end_date_extension


dsajdak (approved):
LGTM!

PR #489, opened 2023-02-09
Signal fixed on changing allocation statuses

rg663:
Resolves issue #474 — Now, allocations that have been given a revoked status display a similar behavior to an allocation that has been marked as denied, sending an email to the user and signaling that the allocation should be disabled. Newlines are also added to email templates to enhance readability.

### DIFF
--- a/coldfront/core/allocation/views.py
+++ b/coldfront/core/allocation/views.py
@@ -232,12 +232,12 @@ class AllocationDetailView(LoginRequiredMixin, UserPassesTestMixin, TemplateView
             if action != 'auto-approve':
                 messages.success(request, 'Allocation Activated!')
 
-        elif old_status != allocation_obj.status.name in ['Denied', 'New']:
+        elif old_status != allocation_obj.status in ['Denied', 'New', 'Revoked']:
             allocation_obj.start_date = None
             allocation_obj.end_date = None
             allocation_obj.save()
 
-            if allocation_obj.status.name == 'Denied':
+            if allocation_obj.status.name == ['Denied', 'Revoked']:
                 allocation_disable.send(
                     sender=self.__class__, allocation_pk=allocation_obj.pk)
                 allocation_users = allocation_obj.allocationuser_set.exclude(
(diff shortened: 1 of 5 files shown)


coldfront/core/allocation/views.py
```diff
@@ -232,22 +232,25 @@ def post(self, request, *args, **kwargs):
             if action != 'auto-approve':
                 messages.success(request, 'Allocation Activated!')
 
-        elif old_status != allocation_obj.status.name in ['Denied', 'New']:
+        elif old_status != allocation_obj.status in ['Denied', 'New', 'Revoked']:
```
aebruno, line 235: This needs to be `elif old_status != allocation_obj.status.name in ['Denied', 'New', 'Revoked']:`

coldfront/core/allocation/views.py
```diff
@@ -232,22 +232,25 @@ def post(self, request, *args, **kwargs):
...
             allocation_obj.start_date = None
             allocation_obj.end_date = None
             allocation_obj.save()
 
-            if allocation_obj.status.name == 'Denied':
+            if allocation_obj.status.name == ['Denied', 'Revoked']:
```
aebruno, line 240: This needs to be `if allocation_obj.status.name in ['Denied', 'Revoked']:`

coldfront/core/allocation/views.py
```diff
@@ -232,22 +232,25 @@ def post(self, request, *args, **kwargs):
...
                 allocation_disable.send(
                     sender=self.__class__, allocation_pk=allocation_obj.pk)
                 allocation_users = allocation_obj.allocationuser_set.exclude(
                                         status__name__in=['Removed', 'Error'])
                 for allocation_user in allocation_users:
                     allocation_remove_user.send(
                         sender=self.__class__, allocation_user_pk=allocation_user.pk)
-
+            if allocation_obj.status == 'Denied':
```
aebruno, line 248: This should be `if allocation_obj.status.name == 'Denied':`

coldfront/core/allocation/views.py
```diff
@@ -232,22 +232,25 @@ def post(self, request, *args, **kwargs):
...
                 send_allocation_customer_email(allocation_obj, 'Allocation Denied', 'email/allocation_denied.txt', domain_url=get_domain_url(self.request))
                 messages.success(request, 'Allocation Denied!')
+            elif allocation_obj.status == 'Revoked':
```
aebruno, line 251: This should be `elif allocation_obj.status.name == 'Revoked':`

aebruno (changes requested):
A few minor fixes..

aebruno (approved):

PR #504, opened 2023-03-02
API for all models documented

rg663:
This PR includes documentation for the following models:

- Allocations
- Fields of Science
- Grants
- Publications
- Research Outputs
- Projects
- Resources
- Users

Documentation includes docstrings for each model's attributes and its methods/properties.

This API documenation can be found in the API tab of the ColdFront documentation.


coldfront/core/allocation/models.py
```diff
@@ -192,6 +234,13 @@ def get_attribute(self, name, expand=True, typed=True,
         return None
 
     def set_usage(self, name, value):
+        """
+        Params:
+            name (str): allocation attribute type whose usage to set
+            value (int): value to set usage to
```
aebruno, line 240: Looks like this should be a float. 

Also remove `# check value type` to get rid of this warning: 
```
WARNING  -  mkdocstrings_handlers: coldfront.core.allocation.models.Allocation.set_usage: Failed to get 'name: description' pair from '# check value type'
```

coldfront/core/resource/models.py
```diff
@@ -53,8 +85,21 @@ def __str__(self):
     class Meta:
         ordering = ['name', ]
 
-
 class Resource(TimeStampedModel):
+    """ A resource is something a center maintains and provides access to for the community. Examples include Budgetstorage, Server, and Software License. 
+    
+    Attributes:
+        parent_resource (Resource) = used for the Cluster Partition resource type as these partitions fall under a main cluster
```
aebruno, line 92: Fix typo.. these should to have a ':' instead of an '='. 

```
WARNING  -  mkdocstrings_handlers: coldfront.core.resource.models.Resource: Failed to get 'name: description' pair from 'parent_resource (Resource) = used for
            the Cluster Partition resource type as these partitions fall under a main cluster'
```

aebruno (changes requested):
This looks great. Had a few minor changes. Also, let's rename some of these files as follows:

1. Let's rename `apidocs/1api.md` to `apidocs/index.md`. This will make that page be the index page.
2. All file names should be lowercase. So let's rename:
    - Fields-of-Science.md
    - Grants.md
    - Publications.md
    - Research-Outputs.md
3. Fix the links in the index.md (1api.md) file to match the new file names.

aebruno (approved):

Issue #495, opened 2023-02-10
Documentation: Project attributes


dsajdak:
Add documentation on the use case examples for new project attributes feature

dsajdak:
#519

Issue #478, opened 2023-01-22
ColdFront requires Python 3.6 which is no longer supported


thomasbergernz:
I'd like to clarify the Requirements which state:
> ColdFront requires Python 3.6.

Does this actually specify the minimum version of Python? Python v3.6 has reached end-of-life...

> As of 2021-12-23, 3.6 has reached the [end-of-life phase](https://devguide.python.org/devcycle/#end-of-life-branches) of its release cycle. 3.6.15 was the final security release. The code base for 3.6 is now frozen and no further updates will be provided nor issues of any kind will be accepted on the bug tracker.

dsajdak:
Yes, thanks for pointing this out.  ColdFront has been tested on the latest version of Ubuntu with latest Python 3.  We'll update this information.

dsajdak:
Resolved with PR #519

PR #509, opened 2023-03-24
Updates and reorganization to documentation

dsajdak:
Addresses issues #495 and #478


dsajdak:
Replacing this with #519 because I messed up some git merges

PR #516, opened 2023-04-11
Fixing Allocations/Resource list incorrect

ShreyasSridhar24:
Fixes issue https://github.com/ubccr/coldfront/issues/467 in which if a resource is marked as not available, it still shows up in the list of resources and as a child resource.
Now, if a resource is marked as not available, it does not show up in the list of resources or as a child resource. The resource still exists but does not show up on the page.
If all the child resources are not available, the child resources list shows that there are no resources available.


dsajdak:
@ShreyasSridhar24 I believe I misspoke when we discussed this.  Only resources that are marked not available should be hidden from the list.  Those marked not allocatable should not be hidden as there are many child resources that aren't allocatable but should be viewed under their parent resource - for example, a compute node might be a child of a cluster partition parent.  The compute node is not allocatable but we want it to show up under the parent resource.  I apologize if I caused this confusion.

dsajdak (changes requested):
@ShreyasSridhar24 I believe I misspoke when we discussed this. Only resources that are marked not available should be hidden from the list. Those marked not allocatable should not be hidden as there are many child resources that aren't allocatable but should be viewed under their parent resource - for example, a compute node might be a child of a cluster partition parent. The compute node is not allocatable but we want it to show up under the parent resource. I apologize if I caused this confusion.

coldfront/core/resource/views.py
```diff
@@ -27,15 +27,22 @@ def test_func(self):
...
 
             {'object': resource,
              'WarrantyExpirationDate': resource.get_attribute('WarrantyExpirationDate'),
              'ServiceEnd': resource.get_attribute('ServiceEnd'),
              'Vendor': resource.get_attribute('Vendor'),
              'SerialNumber': resource.get_attribute('SerialNumber'),
-             'Model': resource.get_attribute('Model'),            
+             'Model': resource.get_attribute('Model'),
+             'Available' : resource.is_available,
+             'Allocatable' : resource.is_allocatable,
```
dsajdak, line 44: I'm not sure if it needs to be changed here as well?

dsajdak (changes requested):
We need to also show non-allocatable child resources on the resource detail page.

Issue #491, opened 2023-02-09
Projects:  Add user enhancement


dsajdak:
Currently, when adding a new user to a project we can search for username or using a name or email address.  However, the only way to add more than one account at a time is by supplying usernames.  It would be good if this could be expanded to include the ability to search and add by email address in bulk.  For our use case, most faculty get student email addresses from their course registry and not having to strip those down to just usernames would be a nicer experience.

PR #521, opened 2023-04-23
Ability to add emails in bulk added

rg663:
Solves issue #491 — you can now add users to a project by providing multiple emails at once. Emails can be separated by spaces or newlines, similar to when searching on multiple usernames.


dsajdak (approved):
Found no issues in local testing

aebruno (changes requested):
When using additional user search classes like ldap user search this fails to filter the results.

Issue #507, opened 2023-03-15
Turn off Grants/Publications/Research Outputs


kayleanelson:
It would be great if there was a way to hide or turn off Grants/Publications/Research Outputs. We aren't going to be launching with those so would like to hide them until we are ready.

PR #524, opened 2023-05-01
Ability to enable/ disable grant, publication, and research output functionality added

rg663:
Resolves issue #507 — users can now toggle whether or not they would like grant, publication, and research output functionality by setting the following environment variables:

- RESEARCH_OUTPUT_ENABLE (True or False)
- GRANT_ENABLE (True or False)
- PUBLICATION_ENABLE (True or False)

To avoid conflicts while reenabling, these apps are installed by default, but if a user sets the environment variable to False for any of the above, it removes all functionality from the frontend.


dsajdak:
This is looking really good.  Providing the option to disable/enable each of these pieces separately is great.  Let's add a few more pieces of logic to this in other areas of ColdFront:

1. On the project detail page at the top are buttons for Add User and Add Publication/Grant/Research Output.  If all three of these are disabled, let's remove this section completely because all that is left is Add User.  If any of them are enabled, we can display the buttons.
2. In the Project Review logic, we force the PI to provide a justification if they have not updated their publications OR grants in the last 365 days.  If these two are both disabled, let's remove this requirement for a justification.  If one or the other is enabled, we can leave the requirement.

dsajdak (approved):
Changes look great.  Functionality works perfect locally but will need to test thoroughly in our development environment.

aebruno:
Closed in favor of #620

PR #528, opened 2023-06-05
Update mokey_oidc readme to include OIDC configuration without Mokey/Hydra integration

cecilialau6776:
I had an issue with using `mozilla-django-oidc` was that the user would log in with OIDC, get redirected to our provider, but ultimately get returned to the login failure url (`/` by default) and not get logged in. The problem was that the `sessionid` cookie was not getting returned to us in Firefox, because the `SESSION_COOKIE_SAMESITE` setting is set to `"Strict"` in ColdFront's default settings in `auth.py`.

I fixed the problem by changing the setting to `"Lax"`.

Also removed an extra newline at the start of the code block.

### DIFF
--- a/coldfront/plugins/mokey_oidc/README.md
+++ b/coldfront/plugins/mokey_oidc/README.md
@@ -20,11 +20,11 @@ Django users.
 - pip install mozilla-django-oidc
 
 ## Usage
+### Mokey/Hydra integration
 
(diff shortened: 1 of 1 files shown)


aebruno (approved):
Thanks for updating the docs. We've had several users ask about this. Much appreciated!

Issue #457, opened 2022-09-23
Project: Auto-enable notifications when role changed


dsajdak:
We do not allow project managers to disable project notifications for their own account.  However, if a user has their notifications disabled and are subsequently changed to a 'manager' role, the notifications are not automatically re-enabled.  They should be.

PR #532, opened 2023-06-08
Notifications are auto-enabled when role is changed from User->Manager

ShreyasSridhar24:
Fixes issue #457 where Notifications were not auto-enabled upon a user being promoted to manager. Now, users who had their notifications disabled will find that their notifications are automatically re-enabled when they become managers.


dsajdak (changes requested):
@ShreyasSridhar24 this works great in my testing.  I think to make this just a tad more user friendly, let's also grey out the notification checkbox for managers on the project user detail page.  When we toggle a user's role to manager and turn on notifications, they can't turn them off so it should be greyed out with a note that managers must have notifications enabled.
  
projectuser

Issue #442, opened 2022-08-01
Grants:  Formatting error


dsajdak:
When a project manager adds a grant and formats the grant amount with commas we see the error 'add a number'  If we remove the commas, it works.  This should be corrected with some logic to accept both formats as well as a dollar sign, if included.

dsajdak:
@aebruno tracked this down to a client side validation issue and may only be seen if you make modifications to the template, like we do.  He recommends we remove this and then it will use the backend validation.

PR #533, opened 2023-06-10
Grants formatting error fixed

rg663:
Resolves issue #442 — Grant amounts in the grant creation/ editing form can now be entered with dollar signs and commas, and percents can be entered with percent signs. In the grant report and project detail pages, grant amounts and percentages are also formatted to be rounded to two decimal places for consistency.

### DIFF
--- a/coldfront/core/grant/forms.py
+++ b/coldfront/core/grant/forms.py
@@ -17,9 +17,9 @@ class GrantForm(ModelForm):
             'direct_funding': 'Direct funding to {}'.format(CENTER_NAME)
         }
         help_texts = {
-            'percent_credit': 'Percent credit as entered in the sponsored projects form for grant submission as financial credit to the department/unit in the credit distribution section. Only digits, spaces, and percent symbols can be entered.',
-            'direct_funding': 'Funds budgeted specifically for {} services, hardware, software, and/or personnel. Only digits, spaces, commas, and dollar signs can be entered.'.format(CENTER_NAME),
-            'total_amount_awarded': 'Only digits, commas, and dollar signs can be entered.'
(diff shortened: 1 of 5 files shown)


coldfront/core/grant/forms.py
```diff
@@ -17,15 +17,15 @@ class Meta:
             'direct_funding': 'Direct funding to {}'.format(CENTER_NAME)
         }
         help_texts = {
-            'percent_credit': 'Percent credit as entered in the sponsored projects form for grant submission as financial credit to the department/unit in the credit distribution section',
-            'direct_funding': 'Funds budgeted specifically for {} services, hardware, software, and/or personnel'.format(CENTER_NAME)
+            'percent_credit': 'Percent credit as entered in the sponsored projects form for grant submission as financial credit to the department/unit in the credit distribution section. Only digits and percent symbols can be entered.',
```
dsajdak, line 20: I'm not seeing this help text.  Is this browser specific?

dsajdak (changes requested):
@rg663 I tried a few different combinations of dumb things to break this and was able to.  Since I'm not seeing the help text I pretended not to know what types of numbers to use in this boxes.  If I enter, for example '5%' in the 'direct funding' box I get an ugly error.  I tried a few other things that would just be accidental typos like "$5%" or "5 000" and got the ugly errors:
`<html>
<body>


Exception Type: | ValueError
-- | --
could not convert string to float: '5%'


</body>
</html>`

It would be good to have nice error messages that ask the user to format correctly.

aebruno (changes requested):
This doesn't seem to work for me. To test I added a grant and entered $1.00 for Direct Funding and $2.00 for Total award amount. After submitting the form the values stored in the database were 100.0 and 200.0 respectively.

aebruno (approved):

Issue #536, opened 2023-06-16
Add ability to add extra links to plugins without modifying ColdFront code


rg663:
Currently, to add extra links from a plugin to ColdFront, a user has to override the ColdFront templates. To make it easier for plugin developers, having a way to add these links without overriding the base code would simplify the process, especially when updating ColdFront versions.

PR #537, opened 2023-06-17
Functionality added to add plugin links to ColdFront without modifying original ColdFront code

rg663:
Resolves issue #536 — A user can now add links to their ColdFront navbar without making changes to the ColdFront core code. This way, if a user updates ColdFront, their core code is untouched by plugin links, making it easy to separate plugins from the base code.

The environment variable is named `EXTRA_PLUGIN_LINKS`, and is a dictionary mapping navbar item names to the urls they map to.

### DIFF
--- a/docs/pages/config.md
+++ b/docs/pages/config.md
@@ -153,7 +153,7 @@ Examples:
 
 Using the example plugin, described in [the plugin creation docs](plugin/how_to_create_a_plugin.md), this is what the environment variable would look like:
 ```
-EXTRA_PLUGIN_LINKS='Weekly Report=weeklyreportapp:weeklyreportapp'
+EXTRA_PLUGIN_LINKS='Weekly Report=weeklyreportapp'
 ```
 
 If your app has the following format in its urls.py file:


coldfront/config/core.py
```diff
@@ -13,6 +13,13 @@
 CENTER_PROJECT_RENEWAL_HELP_URL = ENV.str('CENTER_PROJECT_RENEWAL_HELP_URL', default='')
 CENTER_BASE_URL = ENV.str('CENTER_BASE_URL', default='')
 
+#------------------------------------------------------------------------------
+# Add Plugin Links for Navbar Here
+
+# SETTINGS_EXPORT is a dictionary with keys representing names of urls to show on the navbar, and values that represents the actual urls. The recommended format for the url is "[namespace]:[url_name_inside_namespace]".
+#------------------------------------------------------------------------------
+EXTRA_PLUGIN_LINKS = ENV.dict('EXTRA_PLUGIN_LINKS', cast=eval)
```
aebruno, line 21: Remove the `cast=eval` and be sure to set the default value. Should be:

```
ENV.dict('EXTRA_PLUGIN_LINKS', default={})
```

aebruno (changes requested):

coldfront/templates/common/navbar_plugins.html
```diff
@@ -0,0 +1,5 @@
+{% load common_tags %}
+
+{% for nav_name, url_value in settings.EXTRA_PLUGIN_LINKS.items %}
+    <a class="nav-link" href="{% url url_value %}">{{nav_name}}</a>
```
aebruno, line 4: This doesn't work `{% url url_value %}` so just use the value. Something like:
```
<a class="nav-link" href="{{ url_value }}">{{nav_name}}</a>
```

aebruno (changes requested):

aebruno (changes requested):
This is failing for me with this error: ```UndefinedSettingError at /

"settings.EXTRA_PLUGIN_LINKS" is included in settings.SETTINGS_EXPORT but it does not exist. ```

PR #540, opened 2023-06-20
Allocation Invoice List now includes Allocation Change Requests

### LINKED ISSUE #398: Allocation Changes: Display on invoice menu
We need a revamp of the 'invoice' report.  This only shows allocations in statuses related to payments (payment requested, payment pending).  This doesn't include allocation change requests.  We should discuss the best way to display this.  Perhaps it's just a workflow change where the allocation that the change request is on gets a status change of "payment requested/pending" when the change request requires a payment.  We will need to evaluate what this affects re: slurm and freeipa plugins.

ShreyasSridhar24:
Fixes #398. Allocation Invoice List now includes Allocation Change Requests. These requests now have a payment status, which is a payment-related allocation status for every change request. This enables allocation change requests to appear in the allocation invoice list, and are listed with their payment statuses.


dsajdak (changes requested):
@ShreyasSridhar24 I was getting errors on this one right away so I figured I needed to run a migrate.  This is what I did, please tell me if this is wrong:

```
coldfront makemigrations  
coldfront migrate  
```
It changed alot so that's why I'm a bit skeptical that I did that correctly.  Once I did that I wasn't getting the initial errors I got previously.  However, when I went to add an allocation change request it is broken.  It gives me a red box at the top that just says "payment_status" and will not create the change request.  It should create the allocation change request and set the status to 'pending' which should then display on the allocation change request list.  Once we change the status from pending to something payment related, like "payment requested" then it should show up on the invoice list.

dsajdak (changes requested):
@ShreyasSridhar24 this has some issues.  

1. When a new allocation change request is created it has a default status of paid.  In the current workflow, the status for a new allocation change request is pending.  When approved it gets changed to approved and when denied the status is set to deny.  With your addition of these features, the initial status should still be pending.  However, this status should not be listed at all to the PI user which is what I'm logged in as to create this allocation change request.  

2. This drop down menu should only be available to the admin and should have choices for active, paid, payment pending, payment requested, and denied.  This is because not all allocation change requests involve a cost.  

3. Also, we are now seeing "Status:" listed twice in the Allocation information on the Allocation Change Request page.  This should be updated to represent which status we're referring to.  The top status is Allocation status and the second one is the allocation change status:

alloc_change_status

4. The allocation change request status is not updating properly when changed.  For example, from the allocation change request detail page I change the allocation change status from paid to payment requested.  I view it under the invoice list and rather than updating the existing allocation change request it is creating a new one.  However, it is using the same ID of an allocation change request that already exists (see ID numbers in screenshot):
alloc_change2

Once these are resolved I'll test the workflow further.  Thanks!

PR #548, opened 2023-07-07
 Mention collectstatic for "SITE_STATIC" usage

### LINKED ISSUE #358: Mention collect static for "SITE_STATIC" usage
Just an info for newer users like myself, when running the application through nginx using /static alias in nginx conf, collectstatic will need ran everytime you add/modify custom SITE_STATIC files. The dev runserver will handle this fine without collectstatic but nginx will just result in looking in your nginx conf alias directory and not in SITE_STATIC using django finders

I was going to do a pull request but wasn't entirely sure where this note should be added in the docs. My guess is a note around here? 
https://github.com/ubccr/coldfront/blob/master/docs/pages/config.md?plain=1#L298

cecilialau6776:
Fixes #358.

### DIFF
--- a/docs/pages/config.md
+++ b/docs/pages/config.md
@@ -300,6 +300,12 @@ environment variable:
 SITE_STATIC=/path/to/static/files
 ```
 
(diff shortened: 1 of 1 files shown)


aebruno (approved):
Looks great. Thank You!

Issue #542, opened 2023-06-22
Allocation: Add more complexity to EULA agreements


dsajdak:
We've added some basic updates to allocations with End User License Agreement (EULA) attributes on them in PR #539.  This includes displaying the EULA on the allocation detail page and reminding the manager when new users are added to an allocation that they should be made aware of the EULA.  However, there are some centers that want a more robust process that requires the users on the allocations themselves to login, review, and accept the EULA before they are given access to the resource.  This will involve:

- [ ] changing the user's allocation status to something other than 'active' as they move through this process: if the allocation has a EULA attribute on it, new users that get added to it are in the 'pending' status
- [ ] sending the user an email notifying them that they need to complete this step
- [ ] sending the user an email reminding them after a given time if they don't complete it (should be configurable like the other email features)
- [ ] sending the user an email when they complete the process so they have proof it was done, cc'ing the managers on the project
- [ ] changing the user's allocation status to 'active' once they've completed the EULA review
- [ ] ensure that this does not affect any other allocations that a user might be added to at the same time

There are many possibilities and complications for this type of a process.  This particular feature should provide the expanded capability of requiring a user to agree to a EULA but will not cover all possibilities.  Future users of ColdFront can use this feature to model more complicated processes or expand on this feature, if desired.

aebruno:
@cecilialau6776 Those PRs are quite stale. It looks like #609 by @jrlagrone would be place to pick up and continue with this.

PR #550, opened 2023-07-08
Config options added to make EULA process more thorough

rg663:
Resolves issue #542 — Now, centers who want to have a more thorough process for users on allocations whose resources have a 'eula' attribute can set the config variable `EULA_AGREEMENT` to True. With this, when a user is added to an allocation with a EULA resource, any user that is added (who is not an admin or a PI) has a status of `Pending` instead of `Active`. When this user goes to access the resource, they must either accept or decline the EULA. If they accept, they become an `Active` user on the allocation. However, if they decline, their status is set to `Declined`.

In addition, when a user is added to an allocation with a EULA resource, they receive an email asking them to complete the EULA. When a user agrees to the EULA, an email is sent to the user and their manager to confirm this process. If the config variable `EMAIL_EULA_REMINDERS` is set to True, users are emailed weekly to remind them to complete their EULA process.

### DIFF
--- a/coldfront/core/allocation/views.py
+++ b/coldfront/core/allocation/views.py
@@ -594,6 +594,8 @@ class AllocationCreateView(LoginRequiredMixin, UserPassesTestMixin, FormView):
 
 class AllocationAddUsersView(LoginRequiredMixin, UserPassesTestMixin, TemplateView):
     template_name = 'allocation/allocation_add_users.html'
+    model = Allocation
+    context_object_name = 'allocation'
 
     def test_func(self):
         """ UserPassesTestMixin Tests"""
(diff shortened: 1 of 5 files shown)


coldfront/core/allocation/views.py
```diff
@@ -104,12 +105,38 @@ def get_context_data(self, **kwargs):
...
         attributes_with_usage = [a for a in alloc_attr_set if hasattr(a, 'allocationattributeusage')]
         attributes = alloc_attr_set
 
+
+        allocation_user_active_status_choice = AllocationUserStatusChoice.objects.get(
+                name='Active')
+        if EULA_AGREEMENT:
+            allocation_user_pending_status_choice = AllocationUserStatusChoice.objects.get(name='Pending')
+        
+        allocation_user_status = get_object_or_404(AllocationUser, allocation=allocation_obj, user=self.request.user).status
```
jrlagrone, line 132: This should check all users on an allocation and only show the ability to accept/deny EULA to the relevant user.  As is, this breaks views where the user is not a member of the allocation (e.g. an admin/staff view looking to approve/deny the allocation)

jrlagrone:
We're looking at making similar changes. I'll try to add fixes / updates as we get to them (no eta at this point.)

Right now, this PR is not completely functional (at least for us):

- Views check the current user for EULA approval. This breaks admin/staff/etc views, where the admin/staff/etc is not a member of an allocation
- Approving an allocation changes all users to "Approved" even if they haven't agreed to terms.

PR #556, opened 2023-07-27
Adding additional advanced documentation.

lcrownover:
I had to jump through some hoops to get Coldfront's LDAP authentication working against Active Directory. I believe these extra examples in the documentation could be useful to other institutions who run into the same issues I did.


aebruno (approved):
This looks great. Thanks so much for contributing!

PR #557, opened 2023-07-28
fix misfiring success message, remove redundant function

claire-peters:
Learned earlier today that the post function for the AllocationCreateView was the wrong way to add a success message - it presents the success message even when an error is present on the form. This push corrects that error. It also contains the removal of an unneeded method - the AllocationAddInvoiceNoteView had two get_context_data methods and the second one is the one that would have been in operation, so I deleted the first.

### DIFF
--- a/coldfront/core/allocation/views.py
+++ b/coldfront/core/allocation/views.py
@@ -552,13 +552,9 @@ class AllocationCreateView(LoginRequiredMixin, UserPassesTestMixin, FormView):
         return super().form_valid(form)
 
     def get_success_url(self):
-        return reverse('project-detail', kwargs={'pk': self.kwargs.get('project_pk')})
-
-    def post(self, request, *args, **kwargs):
-        post = super().post(request, *args, **kwargs)
         msg = 'Allocation requested. It will be available once it is approved.'
         messages.success(self.request, msg)
-        return post
+        return reverse('project-detail', kwargs={'pk': self.kwargs.get('project_pk')})
 
 
 class AllocationAddUsersView(LoginRequiredMixin, UserPassesTestMixin, TemplateView):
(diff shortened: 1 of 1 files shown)


coldfront/core/allocation/views.py
```diff
@@ -1234,54 +1230,6 @@ def test_func(self):
         messages.error(self.request, 'You do not have permission to manage invoices.')
         return False
 
-    def get_context_data(self, **kwargs):
```
aebruno, line 1237: I believe we want to keep this function has it contains all the business logic? and delete the other one?

aebruno (changes requested):

claire-peters:
I kept the second because it's the only one that's actually been in use - you can see in [the commit where the first was added](https://github.com/ubccr/coldfront/blob/8e64bfa1ddc8deee870f0baa927c63e095044a9a/coldfront/core/allocation/views.py#L1502-L1550) that it was all added preceding the second get_context_data method, meaning this code has never actually run. I can't really attest to the business logic of the new method for that reason, so I went with the one that's been in use.

That said, I'm happy to instead test the new/first get_context_data function, make sure it works, and remove the older function. Just let me know!

claire-peters:
An additional observation makes the solution even easier - the business logic in the longer get_context_data method is taken straight from AllocationDetailView and has no business being in the AllocationAddInvoiceNoteView class. The second method is definitely the right one.

aebruno:
> An additional observation makes the solution even easier - the business logic in the longer get_context_data method is taken straight from AllocationDetailView and has no business being in the AllocationAddInvoiceNoteView class. The second method is definitely the right one.

@claire-peters Thanks, yes just noticed this as well. Looks good to me!

aebruno (approved):

PR #558, opened 2023-07-31
Add configuration file for default values to populate e.g. status choices for different core applications

claire-peters:
I originally floated the idea of adding a new config file that would serve as a central reference list for all objects that are to be automatically added to the database in functions like add_allocation_defaults [here](https://github.com/ubccr/coldfront/pull/546#discussion_r1256751809). This PR is a follow-up on that. I included all apps with add_default commands. 

There are a few ways I could see this config file being useful for other purposes, e.g., checking whether a given item is in the database in a given view.

### DIFF
--- a/coldfront/config/defaults.py
+++ b/coldfront/config/defaults.py
@@ -1,0 +1,155 @@
+"""Storage of default choices for attribute and status values."""
+
+### Allocation ###
+
+ALLOCATION_DEFAULTS = {
+    'attrtypes': (
+        'Date',
+        'Float',
+        'Int',
+        'Text',
+        'Yes/No',
+        'No',
+        'Attribute Expanded Text'
+    ),
+    'statuschoices': (
+        'Active',
+        'Denied',
+        'Expired',
+        'New',
+        'Paid',
+        'Payment Pending',
+        'Payment Requested',
+        'Payment Declined',
+        'Renewal Requested',
+        'Revoked',
+        'Unpaid',
+    ),
+    'changestatuschoices': ('Pending', 'Approved', 'Denied'),
+    'allocationuserstatuschoices': ('Active', 'Error', 'Removed'),
+    'allocationattrtypes': (
+        # name, attribute_type, has_usage, is_private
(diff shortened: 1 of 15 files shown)


coldfront/config/defaults.py
```diff
@@ -0,0 +1,155 @@
...
+    'attrtypes': (
+        'Date',
+        'Float',
+        'Int',
+        'Text',
+        'Yes/No',
+        'No',
+        'Attribute Expanded Text'
+    ),
+    'statuschoices': (
```
claire-peters, line 15: Was considering adding comments to this file that indicate which default options are required for Coldfront to function - e.g., the "Active" status choice is so ubiquitous in the codebase that removing it as a default would require a lot of changes.

coldfront/config/defaults.py
```diff
@@ -0,0 +1,155 @@
...
+
+### Allocation ###
+
+ALLOCATION_DEFAULTS = {
+    'attrtypes': (
+        'Date',
+        'Float',
+        'Int',
+        'Text',
+        'Yes/No',
```
claire-peters, line 11: Just a side note, but I've been curious about why yes/no is used here instead of booleans and where the "no" option is more appropriate than "yes/no".

coldfront/config/defaults.py
```diff
@@ -0,0 +1,155 @@
...
+### Project ###
+
+PROJECT_DEFAULTS = {
+    'statuschoices': ('New', 'Active', 'Archived'),
+    'projectreviewstatuschoices':  ('Completed', 'Pending'),
+    'projectuserrolechoices': ('User', 'Manager'),
+    'projectuserstatuschoices': (
+        'Active', 'Pending - Add', 'Pending - Remove', 'Denied', 'Removed'
+    ),
+    # in utils.add_project_user_status_choices, projectuserstatuschoices are ['Active', 'Pending Remove', 'Denied', 'Removed'].
```
claire-peters, line 116: Left this question in the code. I didn't find "Pending Remove" in the code anywhere so I think I chose the correct options.

PR #571, opened 2023-09-15
Adding DEFAULT_STATUS and DEFAULT_REMOVE ENV Variables to set users t…

rsandatpsu:
I've added a default status for new users in order for us to use 'Pending - Add' as a default status for any added user to a project.  The default if not added to the coldfront.env is to use 'Active'.  Also for users removed from a project there is a default of 'Pending - Remove' instead of the default of 'Removed'.  If the default is not set in the coldfront.env file it uses 'Removed'.

I had to add code to the project_detail.html file to show users that are in 'Pending - Add' and 'Pending - Remove' status.

### DIFF
--- a/coldfront/config/core.py
+++ b/coldfront/config/core.py
@@ -6,6 +6,13 @@ from coldfront.config.env import ENV
 #------------------------------------------------------------------------------
 
 #------------------------------------------------------------------------------
+# Default User Add and Remove Status
+#------------------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


dsajdak (approved):
This looks good to me.  Thanks @rsandatpsu for your contribution.

To do for @dsajdak  - add this to the [configuration docs](https://coldfront.readthedocs.io/en/latest/config/#coldfront-core-settings)

Issue #572, opened 2023-09-25
Feature: Display a list of slurm accounts a user has on their home page


matthew-kusz:
### Description

I wanted a place where users could quickly see all of the slurm accounts they have without having to look inside their projects and allocations. This list would also include an example of how to use them when submitting a job. I created this list on the users home page and included some additional information like project title and the resources the slurm accounts are for. If this would be a useful feature in ColdFront I can create a PR of the changes I made. It adds some additional code to the slurm plugin. Here's what it looks like with the test data for the cgray user:
![slurm_account_list](https://github.com/ubccr/coldfront/assets/49197515/3af4976a-386d-4194-8c3f-5e61be465b7d)

### Component

Other

dsajdak:
Hi @matthew-kusz!  This would be GREAT to have, in my opinion.  I've got a really old feature request that is similar - see #372.

matthew-kusz:
Sorry for the delay, I've started adding the additional details in #372. We only require the slurm account from coldfront when a user submits a job so I need some additional examples that'll help me figure out how the information should be shown. I'm not sure how the linked resources should be handled and if the information I'm showing is correct for partitions. This is all still from the cgray test data. Here's an example of what I have right now.

From the home page:
![slurm_home_page](https://github.com/ubccr/coldfront/assets/49197515/cf7c3666-c486-477c-89b2-e88968785630)


From an allocation detail page with the University Metered HPC resource:
![slurm_allocation_detail_page](https://github.com/ubccr/coldfront/assets/49197515/15b16bb3-eed4-456d-a9b7-d54ad16f6de3)

Once I have those figured out I need to do some clean up and I'll get a PR set up.

dsajdak:
@matthew-kusz this is amazing!!!  You should not need to specify anything for the linked resources as Slurm handles all of that.  The only thing I'd add is for the University HPC, if the user didn't want to use the cluster default partition and qos they'd have to specify those, like you do for the other clusters but that might be handled by what you already have going here.  This is really great!

Issue #372, opened 2022-02-11
Add Slurm info to allocation detail page


dsajdak:
If the Slurm plug-in is enabled, paint the submission information for a given resource on the allocation detail page.  For example, if we have a partition in the faculty cluster named beta, we'd display: 

To submit to this resource, use the Slurm flags: --cluster=faculty --partition=beta --qos=beta --account=[grabbed from allocation attribute]

This information comes from the resource attributes

PR #589, opened 2023-11-27
Slurm plugin job submission info

matthew-kusz:
This PR adds text to the Slurm plugin telling users what options they need to include when submitting Slurm jobs. More details and some examples can be found in #572, minus the linked resources appearing, and takes care of #372. There are three new env variables added:
1. SLURM_SUBMISSION_INFO - What options should be displayed
2. SLURM_DISPLAY_SHORT_OPTION_NAMES - If the short names for the options should be displayed
3. SLURM_SHORT_OPTION_NAMES - The short name for each option. If an option doesn't have a short name given here then its full name will be used

We only require the Slurm account from ColdFront when a user submits a job, so I relied on the test data for the cgray user to figure out how to show additional options.


dsajdak:
@matthew-kusz that would be great!  Thank you!

PR #596, opened 2024-01-16
Fixed a bug that prevented allocations from being disabled even after it has been denied or revoked

QuanMPhm:
The if statement used the wrong operator to determine if the allocation status was changed to "Denied" or "Revoked". This prevented a signal from being sent that would indicated that the allocation should be disabled.

### DIFF
--- a/coldfront/core/allocation/views.py
+++ b/coldfront/core/allocation/views.py
@@ -238,7 +238,7 @@ class AllocationDetailView(LoginRequiredMixin, UserPassesTestMixin, TemplateView
             allocation_obj.end_date = None
             allocation_obj.save()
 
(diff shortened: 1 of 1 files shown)


aebruno:
@QuanMPhm Apologies for the delay on this. Thanks so much for the bug fix. We now require all commits to be "Signed-off" by using git commit --signoff which will acknowledge that you agree to the [DCO](https://github.com/ubccr/coldfront/blob/main/DCO.md). Can you rebase this commit with --signoff and we'll get this merged. Thanks!

QuanMPhm:
I've added the signoff

aebruno (approved):

PR #612, opened 2024-07-01
ITDEV-34909 - Remove Unused Statuses

peaches7083:
Removed the payment-related statuses from the user-facing list as we are not using them.  I left the underlying models + backend code intact in case we want to use it in the future.  This PR just makes it so end users cannot see statuses we don't want them to use.

https://washu.atlassian.net/browse/ITDEV-34909


peaches7083:
Wrong default branch - closing PR

PR #621, opened 2024-09-20
Fix/local tests

melonistic:
SUMMARY

* added steps for running local tests
* created a requirements-dev.txt to be executed in local development only


melonistic:
Mistaken

PR #425, opened 2022-07-05
[WIP] Simple REST API with OUATH authorization support

mdzik:
Proof-of-concept for rest api. (#384)

For this to work you need to makemigrations for outh2 plugin and install it
 ( pip install django-oauth-toolkit djangorestframework==3.13.1 )

Then run locally, and add 'application': https://django-oauth-toolkit.readthedocs.io/en/latest/getting_started.html#authorization-code

Then copy ClientID and Secret to simple CLI app in Go, from here:
https://github.com/mdzik/coldfront-cli

To just test json output go to

http://localhost:9000/users-api/slurm/RESOURCENAME/

```json
[
    {
        "id": 497,
        "all_public_attributes_as_list": [
            {
                "id": 14, //attribute type ID
                "name": "slurm_account_name",
                "value": "XXXX",
                "usage": ""
            },
            {
                "id": 3,
                "name": "Core Usage (Hours)",
                "value": "1000000",
                "usage": "86048.52"
            }
        ],
        "project_id": 2695
    }
]


```

Issue #316, opened 2021-07-20
Add ColdFront REST API


aebruno:
Add a proper REST API to ColdFront. Features should include:

- Authentication: users should be able to create app credentials that will allow access to only their data. 
- Read access to most data elements: resources, allocations, attributes, grants, pubs, users, etc.
- Write access where appropriate
- Admins should be able to create user accounts?

mdzik:
This independently landed on my TODO list - is anyone working on that? I willing to work on that, but I don't want to duplicate effort :)

aebruno:
@mdzik Would be great to get some help on this and welcome anything you're willing to contribute (code, architecture design suggestions, etc.). At this point I don't think there would be much duplicated effort. We have done some preliminary work on code for authenticating the API (token based auth backend) but it's in the very early stages. @brisco17 is leading the effort on that and may have some code up in the coming weeks. So feel free to hack away on this and we're more than happy to collaborate.

mdzik:
We already use Django REST (https://www.django-rest-framework.org/) framework internally, with its token authentication, for slurm synchronization. It is really mature framework and works like a charm. I would prefer to build in same eco-system, but you probably have some reasons behind not using it?

We could utilize OAuth/OIDC "offline token" mode, as mobile apps are often using. Or JWT-based one. 

Our main goal is to build Go app for, mainly, MOTD for users - stuff like "You have account XX-XXXX at this cluster, with XXX CPUhrs left" or "You need to review your projects" etc. Other use-case is the OnDemand - to provide users drop-down with Accounts to be used in job submission.

I was thinking to start with something simple like (quick draft, rewritten from our internal spec for slurm, sorry for typos etc.):

```
paths:
  /api/allocations/:
    get:
      operationId: listUserAllocations
      description: ''
      parameters: []
      responses:
        '200':
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Allocation'


  /api/allocations/{id}/:
    get:
      operationId: retrieveUserAllocation
      description: ''
      parameters:
      - name: id
        in: path
        required: true
        description: "A unique integer value identifying this Allocation."
        schema:
          type: int
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Allocation'
    post:
      operationId: renevUserAllocation
      description: ''
      parameters:
      - name: id
        in: path
        required: true
        description: "A unique integer value identifying this Allocation."
        schema:
          type: int
[...]

PR #632, opened 2024-11-06
Coldfront REST API

claire-peters:
I built a read-only REST API for FASRC's ColdFront fork and we've been using it internally to query ColdFront data for about half a year. This PR contains the generalizable parts that may help to address issue #316. The views are largely configured to make the same information available to users via API that would be available to them if they were accessing the GUI. PR #425 contains additional configuration for OAuth options etc that may be of use, but I thought having ready-made queries might be useful.


aebruno:
@claire-peters We now require all commits to be "Signed-off" by using git commit --signoff which will acknowledge that you agree to the [DCO](https://github.com/ubccr/coldfront/blob/main/DCO.md). Can you rebase this PR into a single commit with --signoff? Thanks!

claire-peters:
Done! Let me know if any further changes are needed.

aebruno:
@claire-peters This looks great! One question that came up while testing this was about authentication. How are you handling users authenticating to the api at your site? Do you have local user/passwords in ColdFront? 

We're currently using OAuth2/OpenID connect for authentication to ColdFront and it looks like some extract configuration for django-rest is necessary to get this all working together. Currently, when accessing the API there is the default django-rest login page for local user/passwords but that won't work in our case. One option we were exploring was to allow users to login to ColdFront via the web and create an API token on their user profile page. Then use that token for programmatic access to the API using django-rest token based auth. Is this something you have experimented with at all?

claire-peters:
@aebruno We have a custom authentication system that's built on rest_framework token authentication in our ColdFront instance. I realized after you asked that there's an import missing in this PR to enable token authentication to work properly, which isn't missing in ours for that reason - I'll address that.

I think I could extend the userprofile template in the api plugin pretty easily so users can view or regenerate their key there. There's also OAuth config that's part of PR #425 which could be added here. Let me know if you want to go with one or both of those solutions and I can add them!

claire-peters:
Added token-based auth features to the UI. I realize that it may be good to enable or disable said features based on whether token auth is actually enabled or not, but I thought I'd start just with the discussed additions.

matthew-kusz:
I've tested this out a little bit and a couple of things have popped up. I had to pip install django-oauth-toolkit
after the initial install, looks like it's because setup.py wasn't updated to include the package but the requirements text was. The Allocation Request List filter form has two fields marked with [invalid names]. I also see there is a PENDING_ALLOCATION_STATUSES setting, but I don't see it set up anywhere. Everything else looked good.

matthew-kusz:
Also, if I go to the http://localhost:8000/user/user-projects-managers/<username> page and click the button to go back to the user profile page the API Token section disappears. I think this has something to do with it taking me back to the url http://localhost:8000/user/user-profile/<username> rather than just http://localhost:8000/user/user-profile/.

claire-peters:
@matthew-kusz Feedback should all be addressed now, let me know if you catch anything else 👍

claire-peters:
@matthew-kusz All issues found should now be addressed - let me know if you find more!

matthew-kusz:
Thanks, I've done another round of testing and have found a couple more things.
1. Allocation Request List Fulfilled filter set to true always returns nothing
2. There's a Date fulfilled Typo; spelled as fullfilled
3. With the query parameters for projects it doesn't matter if I put true or false, the list of allocations and project users are always included

claire-peters:
@matthew-kusz Fixed all three - let me know if you find anything else!

matthew-kusz:
Tried out the changes and everything seems to be working now.

aebruno:
@claire-peters this is looking good. After you make the above changes, feel free to rebase this PR into a single commit with `--signoff`. Thanks!

matthew-kusz:
Any updates on this?

claire-peters:
@aebruno @matthew-kusz  Changes made and commit squashed - let me know if anything else is needed.

aebruno (approved):

Issue #629, opened 2024-10-29
Feature: ldap search scope


ds-04:
### Description

It appears the ldap search scope is ``one``, it would be useful if it could be ``sub``.

### Component

Other

ds-04:
Certainly auth is using ``onelevel``  via ``python-ldap``, changing to be able to use ``subtree`` would be good.

```
coldfront/config/plugins
diff ldap.py ldap.py.orig

32,33c23,24
< AUTH_LDAP_USER_SEARCH = LDAPSearch(AUTH_LDAP_USER_SEARCH_BASE, ldap.SCOPE_SUBTREE, '(uid=%(user)s)')
< AUTH_LDAP_GROUP_SEARCH = LDAPSearch(AUTH_LDAP_GROUP_SEARCH_BASE, ldap.SCOPE_SUBTREE, '(objectClass=groupOfNames)')
---
> AUTH_LDAP_USER_SEARCH = LDAPSearch(AUTH_LDAP_USER_SEARCH_BASE, ldap.SCOPE_ONELEVEL, '(uid=%(user)s)')
> AUTH_LDAP_GROUP_SEARCH = LDAPSearch(AUTH_LDAP_GROUP_SEARCH_BASE, ldap.SCOPE_ONELEVEL, '(objectClass=groupOfNames)')
```


----------

The ldap user search I think is actually using ``subtree`` by default (``python ldap3`` default) and there is an issue with one of my test accounts in the test OpenLDAP instance in a particular OU, which I'm trying to get to the bottom of.

aebruno:
Agree, this looks good. Probably makes sense to make this a config option?

ds-04:
> Agree, this looks good. Probably makes sense to make this a config option?

Yes it would make sense as a config option.

ds-04:
PR #637 created for review, thanks

ds-04:
was resolved by https://github.com/ubccr/coldfront/pull/641 ... closing this

PR #637, opened 2024-11-15
Update ldap.py with auth search scope subtree capability

ds-04:
Update ldap.py with auth search scope subtree capability


ds-04:
related issue is - Feature: ldap search scope #629

aebruno:
@ds-04 Thanks for the PR. We now require all commits to be "Signed-off" by using `git commit --signoff` which will acknowledge that you agree to the [DCO](https://github.com/ubccr/coldfront/blob/main/DCO.md). Can you rebase this commit with `--signoff`? Thanks!

ds-04:
ok no problem, closing in favour of #641

PR #639, opened 2024-11-20
Implement native plugin mechanism

cc-a:
This PR accompanies #636 and demonstrates a pattern for extending the existing plugin system to prevent the need to edit core coldfront files. The core idea is that by searching for files in the `plugins` directory and using entry points for pip installed packages it's possible to automate the discovery and loading of settings and urlpatterns.

To try it out:
- Checkout this branch.
- Pip install the example plugin I've created at [cc-a/coldfront_native_plugin](https://github.com/cc-a/coldfront_native_plugin) (pip install git+https://github.com/cc-a/coldfront_native_plugin.git).
- Run coldfront showplugins to see the plugin just installed.
- Run coldfront migrate and note that the migration from the plugin is applied.
- Start the development server and open http://localhost:8000/djp-plugin/.

Some thoughts:
- Under the hood this is doing a lot of what djp makes easier to do.
- I've jumped through certain hoops to make sure the existing plugins still work so some complexity .
- Things could be made simpler again by dropping support for loading third party plugins from the `plugins` directory and having developers `pip install -e` them if necessary.

### DIFF
--- a/coldfront/config/settings.py
+++ b/coldfront/config/settings.py
@@ -1,4 +1,7 @@
+from pathlib import Path
+
 import environ
+from importlib.metadata import entry_points
 from split_settings.tools import optional, include
 from coldfront.config.env import ENV, PROJECT_ROOT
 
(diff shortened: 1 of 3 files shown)


aebruno:
I'm probably missing something but this didn't work for me. `coldfront showplugins` gives me an error:
```
$ coldfront showplugins
Unknown command: 'showplugins'
Type 'coldfront help' for usage.
```

Issue #631, opened 2024-11-05
Bug: LDAP user plugin not supporting TLS???


payerle:
### What happened?

I am running a somewhat old version of ColdFront v1.1.0 (I hope to start working on upgrading in the spring, and will be sending a lot of pull requests then) and I need to get TLS working with this version of ColdFront.  I saw that the documentation for the current version mentions a config variable LDAP_USER_SEARCH_USE_TLS seems to do what I want, so I looked at  LDAPUserSearch plugin from the github master in order to backport it to my current version.  However, it looks to me like this is not working, both based on my reading of the code and on some (admittedly hacked) attempts to use the code.

Doing a grep of the latest version of ColdFront, the string LDAP_USER_SEARCH_USE_TLS is only used in 
coldfront/plugins/ldap_user_search/utils.py to set the attribute LDAP_USE_TLS, and that attribute is only used in the utils.py to set the ldap3 Tls instance tls to specify client private key and/or certs to check against when creating the ldap3 Server instance.

In particular, the auto_bind parameter passed to the ldap3 Connection constructor is set to True.  The use of boolean values for the auto_bind parameter of the ldap3 Connection class seems to be somewhat deprecated (at least no longer mentioned in the docs), but looking into the code it is still accepted for backwards compatibility, with the value True being converted to AUTO_BIND_NO_TLS.

I have also copied the LDAPUserPlugin class (from github master) into a script, with some hacks (e.g. replacing the import_from_settings with hard coded values), and running a test case of a lookup, and capturing traffic to the LDAP server, and TLS is not being used when the LDAP_USER_TLS parameter is set to True, but is used when the auto_bind parameter is set to AUTO_BIND_TLS_BEFORE _BIND.

I apologize that I do not currently have the resources to perform a less hacked test, but unless I am missing something big  this does appear to be a mismatch between what the documentation is claiming and what the ColdFront plugin is doing. 

If you confirm the bug, I can provide a pull request which should fix it (as I am developing a fix for my version anyway)

 

### Version

1.1.1

### Component

Other

### What browsers are you seeing the problem on?

None

### Tasks/ user tests when bug is fixed

- [ ]

payerle:
I have verified that the issue is present in ColdFront 1.1.6 (commit  0b96bd391136d0a62ee1840e3cae111e6904b999)  as well.  

I have cloned the latest ColdFront from git, added it to a virtual environment, set .env in site-packages to
PLUGIN_LDAP_USER_SEARCH=True
[...]

PR #643, opened 2024-12-10
Updates to support TLS in ldap_user_search plugin

payerle:
This PR should address concerns of #631

Basically:
1) Adds new config parameter LDAP_USER_SEARCH_CERT_VALIDATE_MODE which is passed to the ldap3.Tls constructor as validate. Accepts as values:
	'required' : Certs are required and must validate
	'optional' : Certs are optional, but must validate if provided
	'none' (or None): Certs are ignored.
The default is None

2) The LDAP_USER_SEARCH_CERT_VALIDATE_MODE is passed as the validate field to the ldap3.Tls constructor

3) If LDAP_USE_TLS is set, we pass the connection parameter 'auto_bind' as ldap3.AUTO_BIND_TLS_BEFORE_BIND instead of simply True Inspection of ldap3 code shows that when this parameter is set to True (a value which is no longer listed in docs as valid) it is treated as AUTO_BIND_NO_TLS, so the previous before of leaving this as True was not doing TLS despite claiming to do TLS. This fix should change that.


payerle:
The above PR should address this issue.  Running under coldfront runserver with same config file, the tcpdump shows the LDAP query as encrypted.  When repeated with bind_dn and passwd, the encryption is starting before the bind.

coldfront/plugins/ldap_user_search/utils.py
```diff
@@ -37,14 +39,26 @@ def __init__(self, user_search_string, search_by):
 
         tls = None
         if self.LDAP_USE_TLS:
+            ldap_cert_validate_mode = ssl.CERT_NONE
+            if self.LDAP_CERT_VALIDATE_MODE == 'none':
```
aebruno, line 43: This first case is not necessary. Change simplify to:
```
            if self.LDAP_CERT_VALIDATE_MODE == 'optional':
                ldap_cert_validate_mode = ssl.CERT_OPTIONAL
            elif self.LDAP_CERT_VALIDATE_MODE == 'required':
                ldap_cert_validate_mode = ssl.CERT_REQUIRED
```

aebruno (changes requested):
@payerle This looks good thanks for the PR. One minor code change requested. Also, we require all commits to be "Signed-off" by using git commit --signoff which will acknowledge that you agree to the [DCO](https://github.com/ubccr/coldfront/blob/main/DCO.md). Can you rebase/add --signoff. Make sure it's a single commit. Thank you!

aebruno:
This has been replaced by #683

PR #645, opened 2025-01-15
OIDC config: set `MOKEY_OIDC_PI_GROUP` from env var

s-sajid-ali:
`MOKEY_OIDC_PI_GROUP` is currently not picked up form the configuration environment variable.

### DIFF
--- a/coldfront/config/plugins/openid.py
+++ b/coldfront/config/plugins/openid.py
@@ -19,6 +19,7 @@ if ENV.bool('PLUGIN_MOKEY', default=False):
     AUTHENTICATION_BACKENDS += [
         'coldfront.plugins.mokey_oidc.auth.OIDCMokeyAuthenticationBackend',
     ]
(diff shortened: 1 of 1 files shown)


s-sajid-ali:
Accidentally closed by deleting branch in fork.

aebruno:
@s-sajid-ali Thanks for the PR. We require all commits to be "Signed-off" by using git commit --signoff which will acknowledge that you agree to the [DCO](https://github.com/ubccr/coldfront/blob/main/DCO.md). Can you rebase/add `--signoff`. Make sure it's a single commit. Thank you!

s-sajid-ali:
Amended the commit. Let me know if that doesn't work.

aebruno:
Looks good thanks. On second thought, I think it makes sense to move this inside the `if ENV.bool('PLUGIN_MOKEY', default=False):` block, as this env variable would only be used if the mokey plugin was enabled. Otherwise this looks good.

s-sajid-ali:
Done.

aebruno:
Looks great, thank you!

PR #646, opened 2025-01-21
Contiguous Internal Project ID; Closes #634

### LINKED ISSUE #634: Feature: contiguous internal project ID
### Description

It would be useful if all projects could be identified in a contiguous fashion. 

~~Perhaps the title prefixed.~~

~~``MyORG-ZZZZ - Textual title which can be changed``~~

~~Where MyORG-ZZZZ is generated by Coldfront, thus all projects titles are created in sequence, and the submitter of the project request can only have the freedom to specify the Textual title, 2nd half of the overall Project title.~~

--------------

**Update:** on reflection, this may be better as an internal project ID field. That is not editable, and is separate from the title field. This field would be viewable by PI/manager/user etc. It would not be editable by anyone. It would increment appropriately.

The internal project ID field, would allow centre staff and users to refer to a common project ID.

```MyORG-ZZZZ ```

As seen in django this would be

| PK  | Internal project ID |
| --- | --- |
| 1 | MyOrg1001 |
| 2 | MyOrg1002 |
| 3 | MyOrg1003 |
| ... | ... |

Beyond the immediate use of an internal project ID field, is later the ability to create LDAP groups based on this field, via a plugin. This is outside of the scope of this initial Issue. Here it is only proposed that an incrementing internal project ID field is created.

### Component

Projects

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by co505**
Closes #634 

**Overview**

Feature implementation of #634, an user defined internal project ID which establishes a one-to-one relationship with the project primary key to improve functionality with LDAP groups. Feature is enabled with the environment variables `PROJECT_CODE` and an optional `PROJECT_CODE_PADDING`. 

**Changes Made**
- Added a new project model class in `models.py`, titled "ProjectCode". A new database table called `project_projectcode` is created to avoid alterations with the preexisting `project_project` table. 
- Adjusted the project `view.py` file to create a `ProjectCode` object upon a `form_valid` response. 
- Adjusted the project `view.py` file to display the defined project code in the `project_detail` and `project_list` templates. 
- Added test cases for the base functionality of the project code feature. 

**Motovation** 

This feature provides a customizable layer for working with the existing project identification system. While the core project ID system remains in place, this enhancement allows HPC departments to customize how they track resource usage and project affiliations and define flexible mappings between projects and LDAP directories.

**Examples** 

If:

`PROJECT_CODE` is set to `EXAMP`
`PROJECT_CODE_PADDING` is set to `3`

Output would be:

`EXAMP001`, `EXAMP002`, `EXAMP003`, `EXAMP004`. 

Each project code will auto-increment alongside the primary key. 

_And_

If `PROJECT_CODE` is changed while in deployment to something like `FEATURE` and `PROJECT_CODE_PADDING` isn't defined in the configuration file (default value is set to `0` to account for this): 

New projects will be saved as `FEATURE3`, `FEATURE4`, in case organisational dynamics change. All preexisting projects will retain the previous project code set to avoid an intrusive reconfiguration risking data integrity. 

**Screenshots**

_Database tables_

image

_Project List View_

image

_Project Detail View_

image

_Coldfront Configuration File_

image

**Future Development to do**

Displaying project code in the administration portal and any other views which may be considered important.

### CODE REVIEW

File: coldfront/config/base.py [hunk 1]
```diff
@@ -18,6 +18,7 @@
 WSGI_APPLICATION = 'coldfront.config.wsgi.application'
 ROOT_URLCONF = 'coldfront.config.urls'
 
+
```
aebruno, line 21: Looks like just whitespace change here, so probably just remove this.

coldfront/core/project/utils.py
```diff
@@ -18,3 +19,6 @@ def add_project_user_status_choices(apps, schema_editor):
 
     for choice in ['Active', 'Pending Remove', 'Denied', 'Removed', ]:
         ProjectUserStatusChoice.objects.get_or_create(name=choice)
+
+
+
```
aebruno, line 24: Just whitespace? remove

coldfront/core/project/models.py
```diff
@@ -15,6 +15,7 @@
 from coldfront.core.utils.common import import_from_settings
 
 PROJECT_ENABLE_PROJECT_REVIEW = import_from_settings('PROJECT_ENABLE_PROJECT_REVIEW', False)
+INSTITUTION_CODE = import_from_settings('INSTITUTION_CODE', False)
```
aebruno, line 18: Where is this initialized in the config? I'm not seeing where `INSTITUTION_CODE` is being used. Did you mean `PROJECT_CODE` here?

aebruno (changes requested):
Thanks for the PR. Here's some initial feedback. 

1. Please remove the erroneous '1' file
2. Let's not include any files that just have whitespace changes
3. Do you envision each project having a single "project code"? If so, let's add this to a column on the project table instead of creating a while new "project_code" object/table. 
4. Any database changes will require a migration which will need to be included in your PR.
5. We require all commits to be "Signed-off" by using git commit --signoff which will acknowledge that you agree to the [DCO](https://github.com/ubccr/coldfront/blob/main/DCO.md). So once your PR is ready, you'll need to rebase into a single commit with `--signoff`

co505:
Many thanks for your response. 

- Any redundant white-spaces and files have been cleaned up. 
- `project_code` is now implemented in to the `Project` class, and therefore as a column into the project table. 
- Views, templates and tests have been altered to fit the adjustment above. 
- Migrations have been included. 
- Commit rebased and signed-off. 

_Worth mentioning:_ 

I've wrote a method for generation of a project_code and placed it inside the project `utils.py` file for easier future debugging/readability.

aebruno (commented):
@co505 This is looking really good. A few comments/suggestions:

There's a few more places in the templates where we list projects that we may want to display the `project_code` instead of the `project_id`? Assuming this is the intended purpose of the using said `project_code`:

Think it is fine to merge as is but just had a quick question in regards to updates. The way this is currently implemented, the `project_code` gets generated once on project creation. So all newly created projects will have a `project_code` generated however any existing projects will be unable to get a `project_code`. One use case we may want to support is the following:  a site has ColdFront installed decides at a later date they want to enable this feature and generate project codes for all existing projects.

A few things to add if we think this use case is important enough to support:

1. Add a management command called `add_project_codes` which loops through all projects that don't have a project code set and generate one. Then sites could run this command at any time to create projects codes via `coldfront add_project_codes`. See a few examples of commands [here](https://github.com/ubccr/coldfront/blob/main/coldfront/core/project/management/commands/add_default_project_choices.py) and [here](https://github.com/ubccr/coldfront/blob/main/coldfront/core/allocation/management/commands/enable_change_requests_globally.py).
2. Should we add some code to generate a `project_code` when updating project information (if one doesn't already exist)? That would go [here](https://github.com/ubccr/coldfront/blob/main/coldfront/core/project/views.py#L484)

co505:
@aebruno Absolutely fair points, I've implemented all of the suggestions. 

Some extra features to mention: 

- I've included a `--dry-run` option into the management command so users can test-run before altering any data.
- In our team, we use Django's administration portal quite often, and having visibility of each project's associated `project_code` directly within the admin interface is beneficial to our workflow. In case other teams have similar workflows, I've implemented a dynamic column into `list_display` with the `get_list_display` function. This shows associated `project_code` inside the main admin interface for `Projects` It's implemented in a somewhat _hacky_ way, by converting the tuple to a list, inserting the column and returning a tuple again. I've tried to minimise any issues by returning the default `list_display` if `PROJECT_CODE` is not set, but if this is something that's not the best to support, I completely understand and I can take it out.

aebruno (approved):

aebruno:
@co505 Looks great, thanks for including the extra features.

PR #648, opened 2025-02-03
Switch to pyproject.toml

pmarr:
This PR converts from setup.py to pyproject.toml and removes setup.py

It enables support of other [pip backends](https://peps.python.org/pep-0517/#terminology-and-goals) besides setuptools like poetry or [uv](https://docs.astral.sh/uv/)

Adds optional dependency groups allowing conditional installs of plugin dependencies.

If interested, I can submit a PR with our production Dockerfile. We build the coldfront project with uv in a minimal, unprivileged container.

Is there any reason upstream would want to stay with setup.py?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ db.sqlite3
 local_data/
 site/
 local_settings.py
+prod_settings.py
 local_strings.py
 coldfront.db
 *.code-workspace
@@ -21,3 +22,5 @@ db.json
 .env
 .devcontainer/*
 .bin/*
+mysql_data/
+static/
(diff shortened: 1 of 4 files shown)


aebruno:
@pmarr Thanks for the PR. We're definitely interested in converting fully to pyproject.toml. This looks like a great start. Testing this now and will provide any feedback. 

>  If interested, I can submit a PR with our production Dockerfile. We build the coldfront project with uv in a minimal, unprivileged container.

This would be great. We're testing out uv now and will likely start using it more in the future. Would be very interested in your Dockerfile if you're willing to share. Thanks!

PR #654, opened 2025-02-27
Fixes ubccr#653

matthew-kusz:
Fixes ubccr#653 by switching to checking if the allocation's status name is in the list rather than equal to it.


matthew-kusz:
Closing in favor of #596.

PR #664, opened 2025-03-24
Add project signals to core/project views.py , signals.py

ds-04:
Add project signals to core/project views.py , signals.py

(required by project_openldap plugin - to push projects to openldap)

### DIFF
--- a/coldfront/core/project/views.py
+++ b/coldfront/core/project/views.py
@@ -33,6 +33,11 @@ from coldfront.core.allocation.models import (Allocation,
                                               AllocationUserStatusChoice)
 from coldfront.core.allocation.signals import (allocation_activate_user,
                                                allocation_remove_user)
+from coldfront.core.project.signals import (project_new,
+                                            project_archive,
+                                            project_activate_user,
+                                            project_remove_user,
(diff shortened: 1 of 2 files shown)


ds-04:
For ``project_{new,archive,update}`` signals, object is sent as multiple fields/attributes are sought (rather than sending pk only)

coldfront/core/project/views.py
```diff
@@ -441,6 +446,10 @@ def post(self, request, *args, **kwargs):
         end_date = datetime.datetime.now()
         project.status = project_status_archive
         project.save()
+
+        # project signals
+        project_archive.send(sender=self.__class__,project_obj=project_obj)
```
aebruno, line 451: This fails with the following error: 
```
    project_archive.send(sender=self.__class__,project_obj=project_obj)
                                                           ^^^^^^^^^^^
NameError: name 'project_obj' is not defined.
```

This should be `project_archive.send(sender=self.__class__,project_obj=project)`

aebruno (changes requested):
Looks good, just need to fix the error.

aebruno (approved):

Issue #659, opened 2025-03-20
Feature: project allocation limit based on allocation type and global per project allocation limit


ds-04:
### Description

Proposed feature: project allocation limit based on allocation type and global per project allocation limit

A proposed feature is to limit the possibility of creating more allocations within a given project. It is envisaged that:
-	The limit would be global (ENV var) and applied per project
-	The limit would be useful to prevent a secondary Compute (e.g. Cluster resource) from being sought – in a scenario where one has already been provided with sufficient access - which may well be the case at a free-at-the-point-of-use centre with a single cluster
-	A mechanism to override on select projects would be possible even within this regime
-	Perhaps even the possibility of not enabling resource requests unless a project is unlocked to do so - though this seems a less important feature

This feature might be useful in a scenario where a centre either auto-allocates or imports the vast majority of projects/allocations.

It will be useful in order to maintain a 1:1:1 mapping of PI:Project:Compute-Allocation.

Secondary:
It would be possible to implement a cap on allocation requests generally, so might be worth enabling at other centres just to provide a hard bound (e.g. with a generous number – 5), to stop any potential sprawl.

### Component

Projects

### Additional information

```shell

```

PR #667, opened 2025-03-27
Adds allocation limits for a resource

matthew-kusz:
This PR adds the ability to set a limit on the number of allocations a resource can have in a project. This is set by creating a new 
`allocation_limit` resource attribute for a resource with the number you want to limit the resource to. It will prevent users from submitting a new allocation request for that resource if the limit is reached. This limit applies to all projects for that resource. This partially handles #659.

allocation_limit


aebruno (approved):
Looks good! Thanks @matthew-kusz

PR #670, opened 2025-04-02
Institution feature implementation. 

co505:
**Overview**

Implemented project institution feature, enabling sites with multiple institutions to easily identify which project belongs to which institution. This feature automates institution assignment based on the PI's email address. 

**Changes Made**

- `models.py` - Added `institution` attribute into the `Project` class.
- `core.py` - Added environment variable, `PROJECT_INSTITUTION_EMAIL_MAP`.
-  `view.py` - Implemented logic into `form_valid` function to handle institution assignment.
-  `/templates` - `project_detail`, `project_list`, `project_archived_list` altered to show `institution` from GUI. 
- `tests.py` - Tests added for `institution` feature.
- `config.md` - Added basic documentation to describe `PROJECT_INSTITUTION_EMAIL_MAP`
- `admin.md` - Overhauled `get_list_display` to add multiple dynamic columns into `list_display`, therefore showing `institution` and `project_code` alongside one another. 
- `/migrations` - Migration file added.
- `/management` - Management commands added to retrospectively assign institution codes to preexisting projects, `--dry-run` option included too.

**Examples***

Feature uses `PROJECT_INSTITUTION_EMAIL_MAP` to check the PIs email address against a dictionary of key-value pairs, set by the user. e.g. `PROJECT_INSTITUTION_EMAIL_MAP=cardiff.ac.uk=CDF,bangor.ac.uk=BGR,swansea.ac.uk=SWAN`

If PIs email matches a key, the value will be stored into the database. 

The screenshots below use a `cardiff.ac.uk` email address and make the `CDF` institution available in the GUI. 

_`project_list`_

image

_`project_detail`_

image

_`Project Admin Panel`_

image


co505:
@aebruno Overhauled this slightly and it's now ready for review.

co505:
@aebruno Rebased PR and updated formatting with `ruff` :)

coldfront/core/project/management/commands/add_institutions.py
```diff
@@ -0,0 +1,69 @@
...
+            self.stdout.write(f"Error: {e}")
+
+    def handle(self, *args, **options):
+        dry_run = options["dry_run"]
+        projects_without_institution = Project.objects.filter(institution="None")
+
+        if dry_run:
+            self.institution_dry_run(projects_without_institution)
+        else:
+            self.update_project_institution(projects_without_institution)
```
Eric-Butcher, line 69: Reduce the duplication of the check for PROJECT_INSTITUTION_EMAIL_MAP and move it into the handler so it only needs to be ran once. 

```suggestion
    def _update_project_institution(self, projects):
        user_input = input(
            "Assign all existing projects with institutions? You can use the --dry-run flag to preview changes first. [y/N] "
        )

        try:
            if user_input == "y" or user_input == "Y":
                for project in projects:
                    project.institution = add_automated_institution_choice(project, PROJECT_INSTITUTION_EMAIL_MAP)
                    project.save(update_fields=["institution"])
                self.stdout.write(f"Updated {projects.count()} projects with institutions.")
            else:
                self.stdout.write("No changes made")
        except Exception as e:
            self.stdout.write(f"Error: {e}")

    def _institution_dry_run(self, projects):
        try:
            for project in projects:
                new_institution = add_automated_institution_choice(project, PROJECT_INSTITUTION_EMAIL_MAP)
                self.stdout.write(
                    f"Project {project.pk}, called {project.title}. Institution would be '{new_institution}'"
                )
        except Exception as e:
            self.stdout.write(f"Error: {e}")

    def handle(self, *args, **options):
        dry_run = options["dry_run"]

        if not PROJECT_INSTITUTION_EMAIL_MAP:
            self.stdout.write(
                "Error, no changes made. Please set PROJECT_INSTITUTION_EMAIL_MAP as a dictionary value inside configuration file."
            )
            return 


        projects_without_institution = Project.objects.filter(institution="None")

        if dry_run:
            self._institution_dry_run(projects_without_institution)
        else:
            self._update_project_institution(projects_without_institution)
```

coldfront/core/project/tests.py
```diff
@@ -276,3 +279,73 @@ def test_different_prefix_padding(self):
         # Test the generated project codes
         self.assertEqual(project_with_code_padding1, "BFO001")
         self.assertEqual(project_with_code_padding2, "BFO002")
+
+
+class TestInstitution(TransactionTestCase):
+    """Tear down database after each run to prevent conflicts across cases"""
```
Eric-Butcher, line 285: I am not sure from looking at this if `TransactionTestCase` is necessary for this test. Could you explain why you are using this over `TestCase`? Or did you just copy-paste from the one above this and it is not needed? I would rather not take the speed hit on tests if it is not necessary, also if this code is not doing anything with transactions it might be confusing to anyone who reads this test and thinks there is more going on than there actually is.

coldfront/core/project/tests.py
```diff
@@ -276,3 +279,73 @@ def test_different_prefix_padding(self):
...
+        self.user.email = "user@bfo.ac.uk"
+        self.user.save()
+        project_institution_two = self.create_project_with_institution("Project 2", PROJECT_INSTITUTION_EMAIL_MAP)
+        self.assertEqual(project_institution_two, "BFO")
+
+        # Create project for user 3
+        self.user.email = "user@inst.edu.com"
+        self.user.save()
+        project_institution_three = self.create_project_with_institution("Project 3", PROJECT_INSTITUTION_EMAIL_MAP)
+        self.assertEqual(project_institution_three, "EDU")
```
Eric-Butcher, line 351: Since `add_automated_institution_choice(project, institution_dict)` (or whatever you rename it to) does not make changes to the database I believe it would be beneficial to add a test case to ensure that that is the case. If someone gets confused down the road and does not realize that and changes the function later, that test should ding the change made to the API for `add_automated_institution_choice`.

coldfront/core/project/models.py
```diff
@@ -106,6 +106,7 @@ def get_by_natural_key(self, title, pi_username):
     history = HistoricalRecords()
     objects = ProjectManager()
     project_code = models.CharField(max_length=10, blank=True)
+    institution = models.CharField(max_length=10, blank=True, default="None")
```
Eric-Butcher, line 109: I'm thinking users might want to put in institution titles that are longer than 10?

Eric-Butcher (commented):
There are a few things that I think would benefit from being refactored but overrall I think this is good!

coldfront/core/project/utils.py
```diff
@@ -48,3 +48,34 @@ def generate_project_code(project_code: str, project_pk: int, padding: int = 0)
...
+    if direct_dict_match:
+        project.institution = direct_dict_match
+        return direct_dict_match
+    else:
+        for key, value in institution_map.items():
+            if key in split_domain[1]:
+                project.institution = value
+                return value
+
+    return project.institution
```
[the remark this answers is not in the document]
co505: @Eric-Butcher My fault with the `IndexError`, I got my wires crossed here. I agree with your suggestion to rename the function too. Many thanks, will commit this as is.

coldfront/core/project/tests.py
co505: @Eric-Butcher Testing isn't my strongest area - I based this on the previous test and welcome the change to `TestCase`. Apologies for this, thanks for the spot.

coldfront/core/project/models.py
co505: @Eric-Butcher I initially set a character limit expecting to use abbreviated institution names, but I've changed that approach and forgot to adjust the limit accordingly. Would 80 characters work better?

co505:
Many thanks again for your thorough review @Eric-Butcher, 

- Suggested title change from `add_automated_institution_choice` to `determine_automated_institution_choice` was implemented, along with other formatting changes to the function. 
- Institution management command suggested changes were implemented too. 
- Test case implemented to confirm  `determine_automated_institution_choice` doesn't change any values in the DB. Along with changes from `TransactionTestCase` to `TestCase`. 
- Institution character limit increased from 10 to 80

coldfront/core/project/models.py
aebruno: Good catch @Eric-Butcher. 80 seems reasonable to me.

coldfront/core/project/migrations/0006_historicalproject_institution_project_institution.py
```diff
@@ -0,0 +1,26 @@
...
+    operations = [
+        migrations.AddField(
+            model_name="historicalproject",
+            name="institution",
+            field=models.CharField(blank=True, default="None", max_length=10),
+        ),
+        migrations.AddField(
+            model_name="project",
+            name="institution",
+            field=models.CharField(blank=True, default="None", max_length=10),
```
aebruno, line 24: Looks like the migration just needs to be updated to reflect the change in length in the model. This should be `max_length=80`

aebruno (changes requested):
@co505 This looks good. Just had one small fix in the migrations for the max_length.

coldfront/core/project/migrations/0006_historicalproject_institution_project_institution.py
aebruno: @co505 Looks good, thanks again for all the great work and best of luck with everything!

aebruno (approved):

PR #609, opened 2024-05-10
Add eula enforcement

jrlagrone:
This is largely based on PR #550 by @rg663 and aims to resolve #542 (and #420?)

Users added to an allocation must agree to the EULA for the associated resource, if it exists.

This adds `PendingEULA` and `DeclinedEULA` statuses to allocation users. An allocation is available to an user if the status of the allocation and the users allocation status are both `Active`. An allocation may be active for some users added to it, but not all added users based on whether or not they accepted the EULA (if applicable). PI's must accept the EULA to submit an allocation request, where enabled.

This attaches EULA agreements to individual allocations, but EULAs are associated with resources. This means an user has to agree to the EULA for each individual allocation on a given resource. 

PIs agree to the EULA when submitting an allocation request.

Some things to consider testing (in addition to those listed in #542):

- [ ] users can only accept a EULA for themselves
- [ ] users choice is reflected in displayed and available allocations
  - [ ] check accounts are generated with appropriate outputs, e.g. try `DEBUG=True PLUGIN_SLURM=True coldfront slurm_dump` and verify that output matches expectations 
  - [ ] Note, there are subtleties -- Users with permissions to see all allocations (like admin or director views) will typically see the status of the allocation. This may not reflect their ability to use the allocation. 
- [ ] users who reject a eula can be removed and re-added to an allocation by a PI/Manager. They will have a new opportunity to accept. Admins can manually reset the check by changing `DeclinedEULA` -> `PendingEULA` status.
- [ ] resources without a EULA should not require agreements
- [ ] setting `EULA_AGREEMENT = False` should disable eula enforcement. Switching that on/off may result in some unexpected behaviors though.
- [ ] **Email reminders have been implemented, but we do not currently have a mail service configured to test.**

PR #671, opened 2025-04-02
Rebased add EULAs

jrlagrone:
Merged and resolved conflicts between main and #609, rebased into a single commit and signed.


cecilialau6776:
@aebruno Just wanted to make sure this PR is on the radar. Thank you!

coldfront/config/core.py
```diff
@@ -25,6 +25,11 @@
 #------------------------------------------------------------------------------
 PROJECT_ENABLE_PROJECT_REVIEW = ENV.bool('PROJECT_ENABLE_PROJECT_REVIEW', default=True)
 
+#------------------------------------------------------------------------------
+# Enable EULA force agreement
+#------------------------------------------------------------------------------
+EULA_AGREEMENT = ENV.bool('EULA_AGREEMENT', default=True)
```
aebruno, line 31: Let's rename this to `ALLOCATION_EULA_ENABLE` and default to False here

coldfront/config/email.py
```diff
@@ -22,3 +22,4 @@
 EMAIL_ALLOCATION_EXPIRING_NOTIFICATION_DAYS = ENV.list('EMAIL_ALLOCATION_EXPIRING_NOTIFICATION_DAYS', cast=int, default=[7, 14, 30])
 EMAIL_SIGNATURE = ENV.str('EMAIL_SIGNATURE', default='', multiline=True)
 EMAIL_ADMINS_ON_ALLOCATION_EXPIRE = ENV.bool('EMAIL_ADMINS_ON_ALLOCATION_EXPIRE', default=False)
+EMAIL_EULA_REMINDERS = ENV.bool('EMAIL_EULA_REMINDERS', default=True)
```
aebruno, line 25: Same here, rename to `EMAIL_ALLOCATION_EULA_REMINDERS` and let's default to False

coldfront/core/allocation/urls.py
```diff
@@ -8,6 +8,8 @@
          allocation_views.AllocationCreateView.as_view(), name='allocation-create'),
     path('<int:pk>/', allocation_views.AllocationDetailView.as_view(),
          name='allocation-detail'),
+     path('<int:pk>/review-eula', allocation_views.AllocationEULAView.as_view(),
```
aebruno, line 11: This URL should only be added if `ALLOCATION_EULA_ENABLE` is True.

coldfront/core/allocation/views.py
```diff
@@ -104,7 +105,18 @@ def get_context_data(self, **kwargs):
         pk = self.kwargs.get('pk')
         allocation_obj = get_object_or_404(Allocation, pk=pk)
         allocation_users = allocation_obj.allocationuser_set.exclude(
-            status__name__in=['Removed']).order_by('user__username')
+            status__name__in=['Removed',]).order_by('user__username')
+        user_in_allocation = allocation_users.filter(user=self.request.user).exists()
```
aebruno, line 109: This block should also probably be guarded by `if ALLOCATION_EULA_ENABLE:`

coldfront/core/project/forms.py
```diff
@@ -45,8 +45,9 @@ class ProjectAddUserForm(forms.Form):
 
 
 class ProjectAddUsersToAllocationForm(forms.Form):
+                
     allocation = forms.MultipleChoiceField(
-        widget=forms.CheckboxSelectMultiple(attrs={'checked': 'checked'}), required=False)
+        widget=forms.CheckboxSelectMultiple(), required=False)
```
aebruno, line 50: Curious why was this removed?

coldfront/core/project/views.py
```diff
@@ -604,6 +612,24 @@ def post(self, request, *args, **kwargs):
                     users_already_in_project.append(ele)
             context['users_already_in_project'] = users_already_in_project
 
+        allocations_with_eula = {}
```
aebruno, line 615: Should we guard here with `if ALLOCATION_EULA_ENABLE:`? it's likely not needed if this is not enabled.

coldfront/core/utils/mail.py
```diff
@@ -125,7 +125,7 @@ def send_allocation_customer_email(allocation_obj, subject, template_name, url_p
     email_receiver_list = []
     for allocation_user in allocation_users:
         if allocation_user.allocation.project.projectuser_set.get(
-                                user=allocation_user.user).enable_notifications:
+                                user=allocation_user.user):
```
aebruno, line 128: Why was this removed? I believe we'll want to keep this?

aebruno (changes requested):
@jrlagrone Thanks again for this PR. Here's a first pass review. Generally, adding a feature like this we prefer to default to having the feature disabled and only if a site wants to enable it should the code modify ColdFront behavior. So we should be able to merge your PR and run ColdFront and nothing should change. Then we enable it and test the new feature.

Also, it would be helpful if you could provide a summary of what this feature does and how we can go about testing it. For example, explain exactly where and what screens are effected when this feature is enabled and what data is required to make it work etc. What a user can expect to see when this is enabled and how to use it. Does this tick all the boxes in #542 ? or a subset? does it add other functionality we should know about? That will make testing much easier. 

Finally, please add any new config options to our documentation [here](https://github.com/ubccr/coldfront/blob/main/docs/pages/config.md#coldfront-core-settings).

coldfront/config/core.py
```diff
@@ -46,6 +50,7 @@
 SETTINGS_EXPORT += [
     'ALLOCATION_ACCOUNT_ENABLED',
     'CENTER_HELP_URL',
+    'EULA_AGREEMENT'
```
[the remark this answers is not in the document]
jrlagrone: This has been changed as requested.

coldfront/config/core.py
jrlagrone: Done

coldfront/config/email.py
jrlagrone: Done

coldfront/core/allocation/urls.py
jrlagrone: Done

coldfront/core/project/forms.py
jrlagrone: Not sure, but I added it back

coldfront/core/project/views.py
jrlagrone: Added guard and removed code that did not appear to be used anywhere

coldfront/core/utils/mail.py
jrlagrone: Re-added. This was an oversight. 

However, it is worth noting that EULA emails should maybe follow a different opt-in/opt-out model than general notifications. It may be worth adding, for instance a `ALLOW_ALLOCATION_EULA_EMAIL_OPT_OUT`  option and associated logic.

jrlagrone:
I'll try to get to the doc updates and adding an option to send verification of acceptance emails as I have time (hopefully by the end of the week)

aebruno:
> I'll try to get to the doc updates and adding an option to send verification of acceptance emails as I have time (hopefully by the end of the week)

@jrlagrone This is great. Thanks so much for the extra details. We'll start reviewing this and get you any additional feedback. Thanks again.

jrlagrone:
@aebruno The docs are updated and added some confirmation email options.

Specifically:

 - Emails can optionally be sent to users when they accept or decline a EULA by setting `EMAIL_ALLOCATION_EULA_CONFIRMATIONS`
    -  Project managers can optionally be cc'd by enabling `EMAIL_ALLOCATION_EULA_CONFIRMATIONS_CC_MANAGERS`
    - User email opt out settings can be ignored for EULA related emails by setting `EMAIL_ALLOCATION_EULA_IGNORE_OPT_OUT`
    - The EULA can optionally be included in the acceptance confirmation emails (if enabled) with `EMAIL_ALLOCATION_EULA_INCLUDE_ACCEPTED_EULA`

Some other notes / potential TODOs:

 - rerunning `add_allocation_defaults` would be necessary if this is turned on for an existing deployment. I believe this is safe because everything is `get_or_create` based at least in that call (whereas rerunning `initial_setup` would likely not be safe.)
 - We've been using this (with other site-specific things attached) and some of are observations are:
   - We have disabled the ability to decline EULAs (they can be accepted or left unanswered). The main reason for this was support burdens -- as it is implemented manual intervention from either a manager (remove and re-add user to allocation) or admin (change user allocation status) is needed to give someone a second opportunity to accept. That would proably be a good config option.
   - There is a desire to conditionally rename "EULA" in some places, e.g. use "Terms and Conditions" or "Usage Guidelines", etc. where appropriate. We've essentially hard-coded these as needed, but it would probably make sense to add a resource attribute like `eula-name` or something to that end for display purposes.
 - I'm unsure how this will behave on allocations with multiple attached resources.

dsajdak:
@jrlagrone THANK YOU for picking up this feature and running with it!  We really appreciate the development and improvements you've made.  I think it will be extremely valuable to many centers.  I was going to wait until I had tested everything to provide feedback but as I'm working my way through testing I'm struggling on a few things and thought I should ask about them now rather than waiting.  The first issue I'm having is that I'm not seeing initial emails go out to users that need to sign the EULA.  How I'm testing is, I create an allocation request for a resource with a EULA attribute. I add a few users to the allocation and see that their allocation status is "PendingEULA" as expected.  However, those users aren't getting an email to tell them to go review the EULA.  Once they login and agree or decline the EULA, those confirmation emails are coming through and all the different env settings you have defined work for me.  I'm just not getting that initial email.  I see that you have a setting to send reminder emails every 7 days and a suggestion to manually force those emails to go out using the django admin interface.  The two scheduled tasks in the django queue are `coldfront.core.allocation.tasks.update_statuses` and `coldfront.core.allocation.tasks.send_expiry_emails` but running those doesn't send out the reminder emails.  Should there be a new task for this feature?  Are the reminder emails separate from the email they should get on allocation creation?

Another thing I wanted to bring to your attention is your logic (or maybe this was ours from our original PR?) that the emails going out shouldn't go to a user that is a PI.  Unfortunately, this may cause problems for collaborative projects where the "PI" of the project requests an allocation for a resource with a EULA and then adds other users to the allocation, some of which may also be PIs.  Is there a different check we can do to get around this?  Something like: don't send the EULA email to the user requesting the allocation.  You can probably tell I'm not a programmer - this may not be possible.

jrlagrone:
@dsajdak I think I see all the issues you pointed out.

> I see that you have a setting to send reminder emails every 7 days and a suggestion to manually force those emails to go out using the django admin interface. The two scheduled tasks in the django queue are ...

That probably needs to be implemented as some sort of migration. It is added here: https://github.com/ubccr/coldfront/blob/527d1cd78a7b6336d3929b4a00cb61e86683892a/coldfront/core/utils/management/commands/add_scheduled_tasks.py#L29-L32 but that would only get called with the initial setup stuff. That's probably mostly documentation, but it is not reasonable (or safe) to rerun the initial setup on a live system.

For the rest, adding users to an allocation (from an allocation page) should send emails with this:

https://github.com/ubccr/coldfront/blob/527d1cd78a7b6336d3929b4a00cb61e86683892a/coldfront/core/allocation/views.py#L811-L821

It looks like it's (potentially) missing from adding users via project pages. I see in our live instance I appear to have just grouped it together with the existing allocation activated and similar emails, so I'll try to untangle that. Maybe the flow should go like (assuming EULAs are enabled):

1. Someone with PI/Manager status submits an allocations request.
    - This user has accepted the EULA in the request
    - (optional) email is sent to requestor confirming acceptance of EULA (**templates exist, but this email is not sent now**)
    - PI / project owner auto-added to the allocation even if they are not the requestor, so they may not have accepted the EULA.
2. PI / Manager adds other users to the allocation.
    - These users need to agree to EULA before they can be 'Active'.
    - Emails about accepting the EULA
        - On being added the the allocation (**may be before allocation is reviewed**) (**optional? -- seems partially implemented -- only catching users from allocation add user form**)
        - On being added to the allocation and the allocation is approved / active (**not implemented -- suggest conditionally adding to the existing allocation activated email?**)
3. Reminders sent 
    - Weekly / configurable reminders (** implemented, but needs instructions / method to enable on existing installs **)

I'll try to make those changes soon.

dsajdak:
@jrlagrone thanks for confirming these things.  I just tried and if I add a new user to an allocation, then I get the email.  If I create a new allocation request and add project users to it, that's when I'm not getting the email.  Good to know.

When I was testing, I was thinking I might not get emails until the allocation is activated.  That would make sense from our use case.  I wouldn't want our users having to login to review the EULA if their allocation wasn't going to get activated.  But I guess other centers might not agree with that.

For allocation renewals, is there a mechanism to force the re-review of the EULA?  It didn't look like it to me so I was going to list that as a future enhancement, but just wanted to make sure I wasn't missing something.  I think our users would hate us if we did that but I think some centers want to force that.

jrlagrone:
@dsajdak That shouldn't be difficult to add re-review for EULAs. 

Locally, we just removed the "Decline Eula" button so users can either accept it or ignore it. Our thought was it doesn't really make sense to let people change, but they still have to accept it. Something like that should be easy to add as an option. 

Alternatively, the status could just stay in "PendingEULA" until it is agreed to. That's maybe a little more subtle, but I think it could function essentially the same way.

I'll look at options for when the EULA emails are sent to.

aebruno (approved):

aebruno:
@jrlagrone merging this now. If there's any follow up changes/fixes feel free to submit a new PR. Thanks again for the work on this feature.

jrlagrone:
> @jrlagrone merging this now. If there's any follow up changes/fixes feel free to submit a new PR. Thanks again for the work on this feature.

It's still on my TODO list to fix the things pointed out by Dori. I don't think it's a large effort, but I still need to find time to do it.

PR #674, opened 2025-04-09
Support Python 3.13

cecilialau6776:
Allows running on Python 3.13+ without the `legacy-cgi` package.

The `cgi` package was deprecated in Python 3.11 and [removed in Python 3.13](https://docs.python.org/3.13/library/cgi.html), which `FormEncode` used [until version 2.1.1](https://github.com/formencode/formencode/releases/tag/2.1.1).

Also removed an unused `pipes` import (which was also removed in Python 3.13).

### DIFF
--- a/coldfront/core/project/views.py
+++ b/coldfront/core/project/views.py
@@ -1,5 +1,4 @@
 import datetime
-from pipes import Template
 import pprint
 import django
 import logging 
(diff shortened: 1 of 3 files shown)


aebruno:
@cecilialau6776 This is fixed in #677

PR #675, opened 2025-04-10
 Add an Allocation Attribute Edit Page

### LINKED ISSUE #293: Dispatch signals for more resource allocation events
As a developer that is developing plugins for ColdFront, I would like my plugin's code to be executed in realtime (or close to) when specific resource allocation events are performed.

This behavior can be implemented by registering methods in the plugin as receivers for particular [Django signals](https://docs.djangoproject.com/en/2.2/topics/signals/).

ColdFront [already dispatches signals for adding and removing a user from a resource allocation](https://github.com/ubccr/coldfront/blob/941e87a6113d108286d8fe439dd4f8c2e619745a/coldfront/core/allocation/signals.py), and [uses this pattern in the FreeIPA plugin](https://github.com/ubccr/coldfront/blob/941e87a6113d108286d8fe439dd4f8c2e619745a/coldfront/plugins/freeipa/signals.py). However, the exposed signals are insufficient for a complete implementation of a generic resource allocation plugin, ex. for OpenStack.

For that, additional signals are required to be dispatched for:
- creation of a new resource allocation
- removal/disabling of a resource allocation
- changing of resource allocation attributes

### LINKED ISSUE #672: Feature: Allow admins to edit Allocation Attributes
### Description

We would like for admins to be able to edit allocation attributes from the allocation detail page. Additionally, I believe that this would help with closing #293 as we would have something to dispatch a signal. I'm happy to work on a PR for this.

### Component

Allocations

### Additional information

```shell

```

cecilialau6776:
This PR adds a page that allows admins to edit allocation attributes and a button on the Allocation Detail page to access this page, which would close #672.

image
image

This PR also adds a signal for allocation attribute changes, which is sent when Allocation Change Requests are approved and through this page, which should [help] close #293.

### DIFF
--- a/coldfront/core/allocation/forms.py
+++ b/coldfront/core/allocation/forms.py
@@ -226,6 +226,24 @@ class AllocationAttributeUpdateForm(forms.Form):
         allocation_attribute.clean()
 
 
+class AllocationAttributeEditForm(forms.Form):
+    attribute_pk = forms.IntegerField(required=False, disabled=True)
+    name = forms.CharField(max_length=150, required=False, disabled=True)
+    orig_value = forms.CharField(max_length=150, required=False, disabled=True)
+    value = forms.CharField(max_length=150, required=False, disabled=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["attribute_pk"].widget = forms.HiddenInput()
+
+    def clean(self):
+        cleaned_data = super().clean()
+        allocation_attribute = AllocationAttribute.objects.get(pk=cleaned_data.get("attribute_pk"))
+
+        allocation_attribute.value = cleaned_data.get("value")
+        allocation_attribute.clean()
+
+
 class AllocationChangeForm(forms.Form):
     EXTENSION_CHOICES = [(0, "No Extension")]
     for choice in ALLOCATION_CHANGE_REQUEST_EXTENSION_DAYS:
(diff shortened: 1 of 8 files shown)


cecilialau6776:
@aebruno, I just wanted to ask if this was on the radar.

aebruno:
@cecilialau6776 Sorry for the delay. Can you rebase this PR to main then we can review. Looks good otherwise, Thanks!

cecilialau6776:
Rebased and added tests!

cecilialau6776:
@aebruno Yup! Squahsed and also rebased.

aebruno (changes requested):
@cecilialau6776 one minor suggestion. Can you remove the "Confirm" button if an allocation has no attributes to edit? For example:



If you click the confirm button, it errors out with a csfr error:



So maybe just have a cancel button to take the user back to previous page? 

Alternatively (or in addition) remove the "Edit Allocation Attributes" button from the allocations pages when no allocation attributes exist, for example here:

cecilialau6776:
@aebruno Thanks, I didn't catch that when I was testing. I've removed the edit button and the confirm button when there are no allocation attributes.
Also noticed a typo for deleting allocation attributes and fixed it :)

cecilialau6776:
Thanks for the feedback! I'd be happy to make the suggested changes. For several of them, I used examples from elsewhere in their respective files, but I can also update those if needed.

Eric-Butcher (commented):
Thanks for making these changes! It all looks great. Also I did not know that there was an HTTPStatus enum, I will have to use that myself!

I don't think it is necessary right now to go and change other code. 

At this stage we are looking to start paying down some of the technical debt in ColdFront, and part of that is making sure new code that comes in is a bit more maintianable than what we had previously. However, the idea was to not be rewriting stuff just yet until we have enough testing to be confidnt that those changes don't break anything.

PR #686, opened 2025-05-14
Add pagination to ColdFront API for large query requests

chrisdaaz:
I love the new ColdFront API! I got an error trying the `/api/projects/?project_users=true` endpoint due to the large number of results. That might have been because I was using SQLite, but maybe this would be useful to have? I wasn't sure what `PAGE_SIZE` to start with. Just a thought!

### DIFF
--- a/coldfront/config/plugins/api.py
+++ b/coldfront/config/plugins/api.py
@@ -20,4 +20,6 @@ REST_FRAMEWORK = {
     'DEFAULT_FILTER_BACKENDS': [
         'django_filters.rest_framework.DjangoFilterBackend'
     ],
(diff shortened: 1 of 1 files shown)


chrisdaaz:
I get an OperationalError: `too many SQL variables` when the `PAGE_SIZE` is too high. I think it's related to SQLite limitations? I'll sign the commit and up the `PAGE_SIZE` to `750`, which works for me.

chrisdaaz:
@aebruno sorry for the delay!

aebruno:
@chrisdaaz Thanks for updating this. Looks like you just need to run `uv run ruff format` to fix the formatting errors.

chrisdaaz:
@aebruno i gave that a try. apologies, i am very unfamiliar with `uv`!

PR #698, opened 2025-06-09
Adds auto_compute_allocation plugin

ds-04:
This PR adds auto_compute_allocation

A README.md is present which covers most of the information required. No management commands are provided.

The automatic allocation is created on project creation.

---

Example non-training project

image

Allocation detail against fictional seahawk cluster, named using project_code

image

Allocation attributes, showing slurm account name based on project_code, institution fairshare, core hours allocated by use of VAR

image


ds-04:
Going to leave the function naming

```
    allocation_auto_compute,
    allocation_auto_compute_attribute_create,
    allocation_auto_compute_fairshare_institute,
    allocation_auto_compute_pi,
```

As it builds upon

-----

Currently I do a lot of work using _Vim_ directly inside a Coldfront Vagrant test instance - which regularly gets teardown/rebuild/reload. Some of the typos have crept in because of this - as I'm not using a full IDE much. There might be other ways of working. Just mentioning for information.

aebruno (approved):
Looks good, thanks @ds-04

PR #699, opened 2025-06-12
Feature/allocation model tests

Eric-Butcher:
This is the start of adding to the Allocation app test suite. I have wrote several tests covering the clean method and the save method. 

Many of the tests for the clean method use `datetime`s. I am not sure how well coldfront currently handles dates and times so these may need to be revised as currently they are quite naive. 

Because the `save()` method uses an external setting to control whether or not it calls arbitrary functions it was quite hard to test and I had to get quite creative. Some of the code for that is admittedly ugly but it seems that it was the only way I could test it without making huge changes to the current way the code is written. I did need to make a small change to the `save()` method so that it refreshes the setting `ALLOCATION_FUNCS_ON_EXPIRE` when the method is called instead of relying on the value from when the module was instantiated by python. Without doing this there would be no way to test that behavior during test time. Some of the code that assists with the `TestCase` for `save()` had to live outside of the actual class because of issues with the way python initializes things and the meta-programming the I had to use to be able to test that an external function actually was being called by save().


Eric-Butcher:
So I went ahead and replaced all occurrences of replacing environment variables using `django.test.override_settings` with monkey patches using `unittest.mock.patch`. 

The problem with testing any functions that relied on the value of environment variables was that the code does not _actually_ use the values of the environment variables inside of its functions. 

For example, `coldfront.core.allocation.models` at the top line where all the import lines are there are lines that create variables with the values of the environment variables:
```
ALLOCATION_ATTRIBUTE_VIEW_LIST = import_from_settings("ALLOCATION_ATTRIBUTE_VIEW_LIST", [])
ALLOCATION_FUNCS_ON_EXPIRE = import_from_settings("ALLOCATION_FUNCS_ON_EXPIRE", [])
ALLOCATION_RESOURCE_ORDERING = import_from_settings("ALLOCATION_RESOURCE_ORDERING", ["-is_allocatable", "name"])
```

Crucially, these are now new variables that are set to the values of the environment variables in the settings, but are not _actually_ the environment variables in the settings. The code in the module then uses these newly created variables. 

This meant that when trying to change the values of the environment variables in test cases using `django.test.override_settings`, the actual variable that the functions were using did not get changed and would be stuck to whatever the default value was in the source code. 

The first approach I used to fix this was to just put a new line at the top of each function that used an env value to get the current value of the environment. However, this was considered to be problematic because it could mean that coldfront would behave differently if a admin ever came in and manually changed the values in settings file at some point in time in the middle of a coldfront deployment. 

The second approach is to use monkey patching instead, which hot-changes the value specifically of the variable inside the particular module when the tests are run using `unittest.mock.patch`. This is a very important distinction to be had because it could cause confusion down the road. **This is not actually changing the value of the environment variable while testing**. It will only change the specific variable in the specific module being tested. You could now get a scenario where you are testing some `module_a.function_a()` that calls inside of it some `module_b.function_b()`. If both of these functions at same point use environment variables, then you would need to perform two separate patches, one for each module. This allows to test without changing the way the code is written. It is just important to be aware of the distinction because it could cause confusion.

Eric-Butcher:
Anton requested that this get split up into different PRs to make it more reviewable. 

Now split into #707  #705 #704 and #703

PR #700, opened 2025-06-18
Feature/uv dockerfile

Eric-Butcher:
This PR aims to replace the old Dockerfile that used python3.8. This new Dockerfile builds the project with uv, supports all the database backends supported by coldfront, and uses a multistage build for faster compose times. 

Users could spin up multiple instances of this container to run their web server,background workers, and migrations.


Eric-Butcher:
REUSE is complaining that there is no license header on the `.dockerignore` which seems odd to me. It is not complaining about there being no license header on the `Dockerfile` however. Is this expected behaviour?

aebruno:
> REUSE is complaining that there is no license header on the `.dockerignore` which seems odd to me. It is not complaining about there being no license header on the `Dockerfile` however. Is this expected behaviour?

You need to add `.dockerignore` to the `REUSE.tom` file. You can put it [here](https://github.com/ubccr/coldfront/blob/main/REUSE.toml#L8) right after the Dockerfile.

PR #711, opened 2025-06-30
Refactored deprecated logging module functionality

Eric-Butcher:
Replaced all instances of logging.WARN with logging.WARNING and logging.warn with logging.warning. This ensures that coldfront will not break due to using deprecated functions which were removed in Python 3.13 and greater.


ANekhai (approved):
Looks good to merge

PR #713, opened 2025-06-30
Refactored django management commands to use self.stdout.write()

Eric-Butcher:
Refactored Django management commands to use `self.stdout.write()` instead of `print()`. This is in line with [Django's docs on writing custom management commands](https://docs.djangoproject.com/en/5.2/howto/custom-management-commands/).


ANekhai (approved):
Changes look good to me

coldfront/core/utils/management/commands/initial_setup.py
```diff
@@ -22,15 +22,17 @@ def handle(self, *args, **options):
             run_setup()
 
         else:
-            print(
-                """WARNING: Running this command initializes the ColdFront database and may modify/delete data in your existing ColdFront database. This command is typically only run once."""
+            self.stdout.write(
+                self.style.WARNING(
+                    """Running this command initializes the ColdFront database and may modify/delete data in your existing ColdFront database. This command is typically only run once."""
```
aebruno, line 27: Can we keep the `"WARNING:"`.. or add `"****WARNING****"`.. would like to keep some indication that this is a potential disastrous operation.

aebruno (approved):

PR #719, opened 2025-07-03
Tests for the Allocation __str__ method.

Eric-Butcher:


ANekhai (approved):
Cases look nice and simple, should be good to merge

Issue #721, opened 2025-07-07
Feature: Random Seed Testing Infrastructure for Reliable Test Runs


Eric-Butcher:
### Description

There currently seems to be a minor issue where certain test cases in coldfront will occasionally fail. This is most likely due to the usage of faker / factories. These create random data for test cases each time the tests are run, but since random values are used each time it is possible that tests only fail in certain specific instances. Right now coldfront is not configured to supply or display the seed used to create these random values, so when tests fail it is not possible to reproduce the exact conditions used for a previous testing run. 

The easiest way to approach this would probably be to simply created a new setting called `TEST_FAKER_SEED` which gets set to a random value on ever test run but can be manually overridden if needed. This could then be used with [factoryboy's reseed method(]https://factoryboy.readthedocs.io/en/stable/reference.html#factory.random.reseed_random). 

A more complicated but potentially more robust solution could be to create a command the inherits from django's built in `manage.py test` command. This could be written to include additional flags for test configuration on the command line. This would most likely eliminate the need for using a setting. 

Alongside this a way to output the seed used with either printing or logging would be useful so that seeds could be recovered for any tests that are ran in case those runs need to be reproduced. 

### Additional information

```shell

```

PR #727, opened 2025-07-15
Set up custom coldfront test runner to set fake seed and stress level.

Eric-Butcher:
This PR includes a custom test runner for coldfront which is a subclass of the default `DiscoverRunner`. This custom test runner adds flags on the command line to specify the seed used for the random value generation with faker as well as an argument to give a stress level to the tests. This custom test runner will also generate and report each time the test suite is ran what the seed used for faker was regardless of whether or user specified it or not. This closes issue #721. 

The only concern with this is that even when specifying a seed manually, there might be an issue if tests are ever run in parallel that might cause inconsistencies. 

Some things that might want to be modified:
1. What to name the file that stores the new test runner code
2. Whether float or int would be preferred for the seed
3. Whether to keep the constants for FAKE_SEED and STRESS_LEVEL as variables imported from the module or to make these settings using django.conf.settings

If you like it as is then it should be ok to merge in.

### DIFF
--- a/coldfront/config/settings.py
+++ b/coldfront/config/settings.py
@@ -8,14 +8,7 @@ from split_settings.tools import include, optional
 from coldfront.config.env import ENV, PROJECT_ROOT
 
 # ColdFront split settings
-coldfront_configs = [
-    "base.py",
-    "database.py",
-    "auth.py",
-    "logging.py",
-    "core.py",
-    "email.py",
-]
+coldfront_configs = ["base.py", "database.py", "auth.py", "logging.py", "core.py", "email.py", "testing.py"]
 
 # ColdFront plugin settings
 plugin_configs = {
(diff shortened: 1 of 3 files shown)


coldfront/config/settings.py
```diff
@@ -8,14 +8,7 @@
...
 # ColdFront split settings
-coldfront_configs = [
-    "base.py",
-    "database.py",
-    "auth.py",
-    "logging.py",
-    "core.py",
-    "email.py",
-]
+coldfront_configs = ["base.py", "database.py", "auth.py", "logging.py", "core.py", "email.py", "testing.py"]
```
ANekhai, line 11: Is there a reason these have been squashed into a single line? I think having each element on a newline makes it easier to read all of them.

ANekhai (approved):
I just have one quick formatting question. The actual content of the test_runner file looks good to merge for me.

coldfront/config/settings.py
Eric-Butcher: I think that is just what ruff did. I can change it back.

Eric-Butcher: Ruff wants it that way. LMK if It is worth a fmt override.

ANekhai: It's probably ok for now then.

PR #737, opened 2025-07-23
add coldfront core setting for PROJECT_UPDATE_FIELDS

### LINKED ISSUE #736: Feature: PROJECT_UPDATE_FIELDS setting
### Description

This setting would change what `Project` attributes PIs/project owners/project managers are able to edit.

Our center uses ColdFront as a tool for managing projects, and we don't want researchers to be able to change project titles. I believe this would be a relatively extensible way to configure what users are able to edit.

### Component

Projects

### Additional information

```shell

```

cecilialau6776:
This allows for center directors to configure which Project fields project managers are able to edit.

Closes #736.

### DIFF
--- a/coldfront/config/core.py
+++ b/coldfront/config/core.py
@@ -126,3 +126,16 @@ PROJECT_CODE_PADDING = ENV.int("PROJECT_CODE_PADDING", default=None)
 # ------------------------------------------------------------------------------
 
 PROJECT_INSTITUTION_EMAIL_MAP = ENV.dict("PROJECT_INSTITUTION_EMAIL_MAP", default={})
+
+# ------------------------------------------------------------------------------
+# Configure Project fields that project managers can update
+# ------------------------------------------------------------------------------
+
+PROJECT_UPDATE_FIELDS = ENV.list(
+    "PROJECT_UPDATE_FIELDS",
+    default=[
+        "title",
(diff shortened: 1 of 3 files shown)


Eric-Butcher:
Tested and this works fine. 

@aebruno 's please look carefully at this one.  I was able to create a weird situation where I was able as a PI to directly change the status of the project to archived without getting a warning first. There might be other odd situations that arise if certain fields are editable like this.

cecilialau6776:
Admittedly I only wanted/expected for this to take away editable fields (for us, title and field of science), and didn't consider the possibility of issues arising from adding other fields.
I think it's worth testing and either disabling or adding a warning about potential ramifications of other fields, but with at least a warning, I think this should be left to the responsibility/discretion of the center director(s)/ColdFront administrators.

cecilialau6776:
I agree that ColdFront doing more validation for directors configuring settings is a much bigger issue.

I'm afraid I don't quite understand what you mean by having certain fields being removed by the setting - to disallow removing certain fields from being edited, or to disallow certain fields from being editable/added to the list?

cecilialau6776:
Ah, I understand. Personally, I think I'd rather have the flexibility as an admin to add fields as well as remove, and I think that the responsibility to configure the app correctly should lie with the directors.

PR #742, opened 2025-07-25
Require Research Output Title

### LINKED ISSUE #738: Bug: Research Output title should be required
### What happened?

Research outputs can be created without a title, and I expect that a title is required.

### Version

1.1.7

### Component

Other

None

### Relevant log output

```shell

```

- [ ]

cecilialau6776:
Closes #738


cecilialau6776:
My bad, missed adding the migration. I was thinking that adding a default title would be acceptable, or populate nulls with just a space (" ").

cecilialau6776:
Not planned for 1.x

PR #751, opened 2025-08-07
Removed '| safe' usages in templates.

Eric-Butcher:
This remediates at least one (and possibly more) minor XSS exploits caused by usage of `| safe` template markings when used inside of script tags. This has been remediated by placing all template-injected JavaScript data into `json_script` template tags and then loading that JSON data inside of scripts using `JSON.parse()`. 

As part of this pull request, markup was removed from `AllocationCreateView` that caused the value of `quantity_label` to be marked as `<strong></strong>` . 

This also introduces a new ruff linter check for "S308" checking improper usage of safe HTML injection into templates: https://docs.astral.sh/ruff/rules/suspicious-mark-safe-usage/. 

Pull request #734 needs to be merged in first. These pull requests address similair issues. This PR will not pass ruff unless #734 is merged.

### DIFF
--- a/coldfront/core/allocation/templates/allocation/allocation_create.html
+++ b/coldfront/core/allocation/templates/allocation/allocation_create.html
@@ -60,11 +60,16 @@ Request New Allocation
   </div>
 </div>
 
+{{ resources_form_default_quantities|json_script:"resources-form-default-quantities" }}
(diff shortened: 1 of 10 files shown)


ANekhai (approved):
I'm less familiar with these front end changes so they could use a second glance. I think they should be ok. The linter check failing is fixed in PR 734.

PR #755, opened 2025-08-14
Added checks for ruff's T10 rule to catch debugging statements.

Eric-Butcher:


ANekhai (approved):
Looks good, I don't think this will break anything and it'd be good to make sure our code is clean of any debugging statements. The branch needs a rebase though.

PR #770, opened 2025-08-19
Replace doi2bib with a util function to remove future package dependency

jrlagrone:
Replace the functionality of the `git_bib` function from `doi2bib`.

This is mainly to address CVE-2025-50817 -- `doi2bib` uses the `future` package and `doi2bib` appears to no longer be actively maintained. It appears to me that the functionality being used in ColdFront is a simple url request to the crossref.org rest api so just re-implement that.

Note: the following have not been updated (but should be).

- the tests in publication will no longer work as written (I didn't fully comprehend what it was actually doing, but I can circle back when I have more time available and try to update them)
- The packages `doi2bib` and `future` should no longer be dependencies in the `uv` file. I'm still running an older version using `setup.py` / `requirements.txt`, so I didn't test. My assumption is that is a trivial change.


Eric-Butcher:
The boolean return value is never actually used anywhere this is called, so overall it probably makes since to change the API of this function. However the bibtex library seems to be going from v1 to v2 and a lot of the view code is going to have to be refactored anyways as some point. So I argue that it should all just be left for another day. 

This is fine for now.

Eric-Butcher:
@jrlagrone thanks for catching this!

Eric-Butcher:
@jrlagrone Make sure to squash and sign the commit when you are ready for his pr to be merged.

aebruno (changes requested):
> This is mainly to address [CVE-2025-50817](https://github.com/advisories/GHSA-xqrq-4mgf-ff32)

This CVE has been withdrawn and determined not to be a security flaw in python-future.

> doi2bib appears to no longer be actively maintained.

What if we switch to another module that is actively maintained? I suggest we either look at [crossrefapi](https://github.com/fabiobatalha/crossrefapi) or [habanero](https://github.com/sckott/habanero). Both of these appear actively maintained with good APIs. This way we can likely simplify this one step further and get rid of the BibTexParser code as well. 

> It did change the uv file more than I expected, but it looks like that's mostly upload-time fields so I assume that's not really an issue. I'm pretty new to UV though, so it might be good to look at that closely.

The changes to the uv.lock file are likely due to using a newer version of uv which added `upload-time` to the lock file. This also bumps the lock file version. I'll be making these changes in a separate PR along with adding a `tool.uv.required-version`. 

In summary, here's some initial feedback on this PR:

1. Let's modify the `def get_bib(doi_number):` function to use one of the modules suggested above. The function should just return the parsed python dict with all the publication data parsed or raise an exception on error. This way callers will have all the data they need without having to parse bibtex. 
2. This may require some more code changes to the bits that use the parsed data structure as it might be slightly different but shouldn't be too bad as all the same data is there.
3. Remove the uv.lock changes as those will be included in another PR.

jrlagrone:
I'm good with that. It might be a while before I can work on it -- I can try to prioritize it if you want to merge it before migrating repos. 

Otherwise, I think I'm good closing or changing this to WIP for now and including it as a recommended improvement / refactor for "later". Currently nothing is broken.

PR #771, opened 2025-08-19
ruff check for UP. Fixed code to comply with the UP linter rules.

Eric-Butcher:
This was also a bit of a doozy to implement, but it will go a long way to assuage some of the concerns that people have brought up about ColdFront have a more opinionated and consistent style going forward in the code.

Most of this was just updating things to use f-strings. Much more readable, faster, and more modern. 

This will have to be merged after #734 and #751.


Eric-Butcher:
Seems that after all the new work that is gone on in the repo when I try to rebase this something is getting messed up. Fails on py3.10 but succeeds on 3.12, I will have to look into this further.

PR #772, opened 2025-08-20
add project_openldap to docs

simonLeary42:
Feel free to change the wording.


ds-04:
@simonLeary42 

Current wording:

```
- LDAP User Search (read only)
- LDAP Projects (read/write)
```


Would be tempted to change wording to:

```
- LDAP protocol User Search
- Project OpenLDAP
```

The other link descriptions don't describe the plugins, though i think it makes sense that user search include the word protocol, as its somewhat universal.

PR #782, opened 2025-08-28
Add support for slurm parent accounts.

### LINKED ISSUE #746: Feature: Slurm Plugin: Parent accounts
### Description

As of right now, I don't see a way for the Slurm plugin to support assigning a parent account to an account. I propose adding a `slurm_parent_account` attribute and updating the slurm plugin appropriately.

### Component

Other

### Additional information

```shell

```

cecilialau6776:
Adds support for Slurm allocations to have parents/children.
- added tests for `associations.py`
- adds a new `slurm_children` allocation attribute
- updated slurm plugin readme

I could also add an admin-only notification for invalid `slurm_children` (i.e. when a `slurm_children` account doesn't exist)

Closes #746.


aebruno:
@cecilialau6776 Thanks for the PR. Can you confirm this doesn't change the existing behavior in any way. i.e. this only adds functionality? Also, can you squash into single commit. Thanks!

cecilialau6776:
I'm looking this over again - I think I'd like to shift "setting account children" to "setting account parent". Unfortunately, I don't quite remember why I did it this way...

PR #785, opened 2025-09-05
DOCS: New page on how to upgrade coldfront

chrisdaaz:
I made a page for the documentation site based on [my notes](https://gist.github.com/chrisdaaz/31400beff47dbdbf33f205b2fee22b47) from upgrading my coldfront codebase to the latest changes. 

cc: @ds-04


aebruno:
@chrisdaaz Thanks for the PR. This will be a great addition to our docs. We have an existing section for upgrading ColdFront [here](https://docs.coldfront.dev/en/stable/upgrading/) and think it might make sense to add your notes to this page. Suggest we add what you have to a `Upgrading with Git` section. Can you move this into that page. Thanks!

chrisdaaz:
@aebruno sorry for the delay -- thanks for the feedback!

aebruno:
> @aebruno sorry for the delay -- thanks for the feedback!

No prob! Can you also squash to single commit. Thanks

PR #786, opened 2025-09-05
add local_urls to match local_settings

simonLeary42:
Please correct me if I'm wrong:

A typical django workflow involves creating a project with `django-admin startproject`, then you get a "project directory", which is distinct from the "app directory". A django "project directory" contains `manage.py`, `settings.py`, `urls.py`, `asgi.py` , and `wsgi.py`.

Coldfront does away with the project directory and instead:
* `manage.py` is replaced with the `coldfront` binary
* `settings.py` is replaced with `coldfront/config/settings.py` which the user can extend
* `wsgi.py` is replaced with `coldfront/config/wsgi.py`
* `asgi.py` is removed
* `urls.py` is replaced with `coldfront/config/urls.py` which the user cannot extend without keeping a dirty worktree

I would like to extend the URLs config without keeping a dirty worktree, so I added `local_urls.py` which behaves exactly like `local_settings.py`.

### DIFF
--- a/coldfront/config/urls.py
+++ b/coldfront/config/urls.py
@@ -6,6 +6,8 @@
 ColdFront URL Configuration
 """
 
+import environ
+import split_settings
 from django.conf import settings
 from django.contrib import admin
 from django.core import serializers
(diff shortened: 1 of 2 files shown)


aebruno (changes requested):
@simonLeary42 Can you fix up the linting errors. Thanks!

simonLeary42:
For the record, the reason I want to do this is so I can use https://github.com/django-commons/django-debug-toolbar

PR #795, opened 2025-09-16
Add tagging system to projects/allocations/resources

cecilialau6776:
The `Tag` model is built to be added as a ManyToMany field onto any arbitrary model. This PR adds it to `Project`, `Allocation` and `Resource`.

The `allowed_models` field dictates which tags show up as part of editing in the UI - this is **not** validated anywhere else. An admin could add a tag to a model that's not in the list through the admin panel or the shell.

The project detail and allocation detail pages have a card for tags, while resources can only have their tags edited via the admin panel.

Permissions on tags can be set such that:
- Anyone can view
- Project managers can add/remove the tag
- Only superusers can view/add/remove

Only administrators/superusers can create tags, and only through the admin portal.

Potential future features:
- searching objects (projects/allocations) by tag
- more robust permissions

### DIFF
--- a/coldfront/core/allocation/tests/test_models.py
+++ b/coldfront/core/allocation/tests/test_models.py
@@ -61,7 +61,7 @@ class AllocationModelCleanMethodTests(TestCase):
 
     def test_status_is_expired_and_end_date_not_past_has_validation_error(self):
         """Test that an allocation with status 'expired' and end date in the future raises a validation error."""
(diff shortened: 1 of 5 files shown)


simonLeary42:
After doing a deep dive on allocation statuses, I came to the conclusion that most of the statuses didn't really belong as statuses. They belong as tags. When a PI requests an allocation renewal, the allocation shouldn't stop being `Active` and start being `Renewal Requested`, it should stay as `Active` and just get an extra tag.

PR #796, opened 2025-09-16
Added setting FIELD_OF_SCIENCE_HIDE 

csbarnet:
This PR adds the setting FIELD_OF_SCIENCE_HIDE that removes or hides references to Field of Science from all pages when True. It defaults to False, so that the default behavior of Coldfront does not change. It does not _remove_ the field of science field from the Project model.

In many cases, references to the field of science were removed by simply adding conditional inclusion of the field on respective templates. This includes the allocations by field of science graphic on the center-summary page. Additionally, references to relevant fields in javascript are also handled.

In the cases of the ProjectSearchForm and the ProjectCreationForm (renamed ProjectUpdateForm), the django crispy-forms FormHelper object was used to dynamically include/exclude the Field of Science search field. 

The ProjectUpdateView was assigned the same form_class as the ProjectCreateView so that they share fields.

Tests test inclusion or exclusion of the Field of Science depending on the setting using the @override_settings decorator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@ db.json
 .env
 .devcontainer/*
 .bin/*
+local_resources/*
+
+coldfront/plugins/sftocf/*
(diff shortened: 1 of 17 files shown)


csbarnet:
this has some extra cruft in it. will resubmit

PR #807, opened 2025-10-02
project_openldap: make function names more generic

simonLeary42:
Many of the functions in `project_openldap.utils` are generic for any LDAP entry, with unnecessarily specific names. I want to import these functions and use them for `allocation_ldap`.

All these renames were done with vscode/pylance [rename symbol](https://code.visualstudio.com/docs/editing/refactoring#_rename-symbol) feature, which promises to fix all references to a symbol across the entire codebase.

other changes:
* `update_project_posixgroup_in_openldap` -> `update_project_posixgroup_description_in_openldap` is actually *less* generic


simonLeary42:
From @ds-04 (author of `project_openldap`) on Slack:

> this is much needed, and i thought about it doing it .. but it got to the point where i just needed the code in core

> only comment is perhaps .. do we use posixGroup in the name .. or posixgroup

> e.g. OpenLDAP terms in the functions names .. and their caseing

I dont care much about the capitalization, feel free to suggest edits on my PR and I will apply them

> yeah im not that bothered either

> its fine i think

> ive not tested it .. but it looked fine

PR #809, opened 2025-10-02
project_openldap: add config docs

simonLeary42:
All the other plugins have their configs in this file, so `project_openldap` should have its config here too.


simonLeary42:
superceded by https://github.com/ubccr/coldfront/pull/818

PR #816, opened 2025-10-03
refactor allocation status lists

simonLeary42:
[The list of allocation statuses](https://docs.coldfront.dev/en/stable/howto/allocations/statuses/) is long and rather confusing. Coldfront's behavior with respect to these statuses is inconsistent. It seems as though the list of statuses has grown over time, and some statuses were added without also being accounted for in various filters and conditionals. This isn't surprising because accounting for all statuses in all filters and conditionals is a lot of work. I think many of these statuses would be more appropriate as [tags](https://github.com/ubccr/coldfront/pull/795).

I have a [discussion post about allocation statuses](https://github.com/ubccr/coldfront/discussions/859) where I describe how I think they are supposed to work.

Assuming the above rules are followed, Coldfront's behavior is still surprising in many situations:
* If an allocation changes from status `Active` to status `Expired`, the `allocation_disable` and `allocation_remove_user` signals will not be triggered, so plugins cannot respond. These signals trigger only when an allocation changes to statuses `Denied` or `Revoked`.
* If a PI removes a user from their project, that user will be removed from all their project's allocations, except allocations with status `Expired` or `Revoked`.
* Allocations with statuses `New`, `Paid`, `Payment Pending`, and `Payment Requested` count towards the allocation limit, but allocations with statuses `Approved`, `Payment Declined`, `Pending`, and `Unpaid` do not.
* The start date cannot be removed from an allocation with status `Active`, but it can be removed from an allocation with status `Renewal Requested`.
* The end date cannot be removed from an allocation with status `Expired`, but it can be removed from an allocation with status `Revoked`.

This PR does not change this behavior in any way, instead it organizes the behavior for all situations in one file, so it can be easily understood. This PR also makes this behavior configurable, which is a nice bonus.

The tests all pass. For reference, here are the tests that fail if I make all the allocation status lists into empty lists:

<details>

```
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
FFF..FFF................F......FFF............F.............................................................s..............sssss...
======================================================================
FAIL: test_status_is_active_and_no_end_date_has_validation_error (coldfront.core.allocation.tests.test_models.AllocationModelCleanMethodTests.test_status_is_active_and_no_end_date_has_validation_error)
Test that an allocation with status 'active' and no end date raises a validation error.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_models.py", line 114, in test_status_is_active_and_no_end_date_has_validation_error
    with self.assertRaises(ValidationError):
AssertionError: ValidationError not raised

======================================================================
FAIL: test_status_is_active_and_no_start_date_has_validation_error (coldfront.core.allocation.tests.test_models.AllocationModelCleanMethodTests.test_status_is_active_and_no_start_date_has_validation_error)
Test that an allocation with status 'active' and no start date raises a validation error.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_models.py", line 106, in test_status_is_active_and_no_start_date_has_validation_error
    with self.assertRaises(ValidationError):
AssertionError: ValidationError not raised

======================================================================
FAIL: test_status_is_active_and_start_date_after_end_date_has_validation_error (coldfront.core.allocation.tests.test_models.AllocationModelCleanMethodTests.test_status_is_active_and_start_date_after_end_date_has_validation_error)
Test that an allocation with status 'active' and start date after end date raises a validation error.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_models.py", line 125, in test_status_is_active_and_start_date_after_end_date_has_validation_error
    with self.assertRaises(ValidationError):
AssertionError: ValidationError not raised

======================================================================
FAIL: test_status_is_expired_and_end_date_not_past_has_validation_error (coldfront.core.allocation.tests.test_models.AllocationModelCleanMethodTests.test_status_is_expired_and_end_date_not_past_has_validation_error)
Test that an allocation with status 'expired' and end date in the future raises a validation error.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_models.py", line 68, in test_status_is_expired_and_end_date_not_past_has_validation_error
    with self.assertRaises(ValidationError):
AssertionError: ValidationError not raised

======================================================================
FAIL: test_status_is_expired_and_no_end_date_has_validation_error (coldfront.core.allocation.tests.test_models.AllocationModelCleanMethodTests.test_status_is_expired_and_no_end_date_has_validation_error)
Test that an allocation with status 'expired' and no end date raises a validation error.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_models.py", line 59, in test_status_is_expired_and_no_end_date_has_validation_error
    with self.assertRaises(ValidationError):
AssertionError: ValidationError not raised

======================================================================
FAIL: test_status_is_expired_and_start_date_after_end_date_has_validation_error (coldfront.core.allocation.tests.test_models.AllocationModelCleanMethodTests.test_status_is_expired_and_start_date_after_end_date_has_validation_error)
Test that an allocation with status 'expired' and start date after end date raises a validation error.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_models.py", line 79, in test_status_is_expired_and_start_date_after_end_date_has_validation_error
    with self.assertRaises(ValidationError):
AssertionError: ValidationError not raised

======================================================================
FAIL: test_allocationaddusersview_access (coldfront.core.allocation.tests.test_views.AllocationAddUsersViewTest.test_allocationaddusersview_access)
Test access to AllocationAddUsersView
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 332, in test_allocationaddusersview_access
    self.allocation_access_tstbase(self.url)
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 69, in allocation_access_tstbase
    utils.test_logged_out_redirect_to_login(self, url)
  File "/srv/simon/src/coldfront/coldfront/core/test_helpers/utils.py", line 40, in test_logged_out_redirect_to_login
    test_case.assertRedirects(response, f"/user/login?next={page}")
  File "/srv/simon/src/coldfront/.venv/lib/python3.12/site-packages/django/test/testcases.py", line 555, in assertRedirects
    self.assertEqual(
AssertionError: 302 != 200 : Couldn't retrieve redirection page '/allocation/1/': response code was 302 (expected 200)

======================================================================
FAIL: test_allocationchangeview_access (coldfront.core.allocation.tests.test_views.AllocationChangeViewTest.test_allocationchangeview_access)
Test get request
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 172, in test_allocationchangeview_access
    self.allocation_access_tstbase(self.url)
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 69, in allocation_access_tstbase
    utils.test_logged_out_redirect_to_login(self, url)
  File "/srv/simon/src/coldfront/coldfront/core/test_helpers/utils.py", line 40, in test_logged_out_redirect_to_login
    test_case.assertRedirects(response, f"/user/login?next={page}")
  File "/srv/simon/src/coldfront/.venv/lib/python3.12/site-packages/django/test/testcases.py", line 555, in assertRedirects
    self.assertEqual(
AssertionError: 302 != 200 : Couldn't retrieve redirection page '/allocation/1/': response code was 302 (expected 200)

======================================================================
FAIL: test_allocationchangeview_post_extension (coldfront.core.allocation.tests.test_views.AllocationChangeViewTest.test_allocationchangeview_post_extension)
Test post request to extend end date
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 183, in test_allocationchangeview_post_extension
    self.assertContains(response, "Allocation change request successfully submitted.")
  File "/srv/simon/src/coldfront/.venv/lib/python3.12/site-packages/django/test/testcases.py", line 660, in assertContains
    self.assertTrue(
AssertionError: False is not true : Couldn't find 'Allocation change request successfully submitted.' in response

======================================================================
FAIL: test_allocationchangeview_post_no_change (coldfront.core.allocation.tests.test_views.AllocationChangeViewTest.test_allocationchangeview_post_no_change)
Post request with no change should not go through
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 193, in test_allocationchangeview_post_no_change
    self.assertContains(response, "You must request a change")
  File "/srv/simon/src/coldfront/.venv/lib/python3.12/site-packages/django/test/testcases.py", line 660, in assertContains
    self.assertTrue(
AssertionError: False is not true : Couldn't find 'You must request a change' in response

======================================================================
FAIL: test_allocationremoveusersview_access (coldfront.core.allocation.tests.test_views.AllocationRemoveUsersViewTest.test_allocationremoveusersview_access)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 355, in test_allocationremoveusersview_access
    self.allocation_access_tstbase(self.url)
  File "/srv/simon/src/coldfront/coldfront/core/allocation/tests/test_views.py", line 69, in allocation_access_tstbase
    utils.test_logged_out_redirect_to_login(self, url)
  File "/srv/simon/src/coldfront/coldfront/core/test_helpers/utils.py", line 40, in test_logged_out_redirect_to_login
    test_case.assertRedirects(response, f"/user/login?next={page}")
  File "/srv/simon/src/coldfront/.venv/lib/python3.12/site-packages/django/test/testcases.py", line 555, in assertRedirects
    self.assertEqual(
AssertionError: 302 != 200 : Couldn't retrieve redirection page '/allocation/1/': response code was 302 (expected 200)

----------------------------------------------------------------------
Ran 131 tests in 3.297s

FAILED (failures=11, skipped=6)
Destroying test database for alias 'default'...
```

</details>

Future work:
* a few `FIXME` comments have been added, these should be fixed as soon as this PR goes through
    * I would just fix them in this PR, but I want the scope of this PR to be limited to purely refactoring, no behavior changes
* many of these status lists should be combined and simplified
    * I would personally like to see the status list reduced to just `New`, `Active`, `Inactive`, with all other statuses becoming tags
* some of the above behavior should probably be changed
* if I am right about how these statuses were meant to be used, I would like to add it to the documentation, and I would like to see enforcement that they are used this way
* `ALLOCATION_STATUSES_UNSET_START_DATE_END_DATE` and `ALLOCATION_STATUSES_DO_DISABLE` should be separate conditionals, rather than having one nested in the other
* `ALLOCATION_STATUSES_DO_REMOVE_USER` and `ALLOCATION_STATUSES_DO_REMOVE_USER_RENEW` should not both exist, the same list should be used in both places. (https://github.com/coldfront/coldfront/issues/857#issuecomment-3457839085)

### DIFF
--- a/coldfront/config/core.py
+++ b/coldfront/config/core.py
@@ -55,11 +55,124 @@ ALLOCATION_DEFAULT_ALLOCATION_LENGTH = ENV.int("ALLOCATION_DEFAULT_ALLOCATION_LE
 ALLOCATION_ACCOUNT_ENABLED = ENV.bool("ALLOCATION_ACCOUNT_ENABLED", default=False)
 ALLOCATION_ACCOUNT_MAPPING = ENV.dict("ALLOCATION_ACCOUNT_MAPPING", default={})
 
+# ------------------------------------------------------------------------------
+# Lists of statuses by which to select allocations for various situations
+# see docs/pages/config.md for more information
+# ------------------------------------------------------------------------------
+ALLOCATION_STATUSES_ALLOW_RENEW = ENV.list("ALLOCATION_STATUSES_ALLOW_RENEW", default=["Active"])
+ALLOCATION_STATUSES_DO_ACTIVATE = ENV.list("ALLOCATION_STATUSES_DO_ACTIVATE", default=["Active"])
+ALLOCATION_STATUSES_REQUIRE_EULA = ENV.list("ALLOCATION_STATUSES_REQUIRE_EULA", default=["Active"])
+ALLOCATION_STATUSES_REQUIRE_START_DATE = ENV.list("ALLOCATION_STATUSES_REQUIRE_START_DATE", default=["Active"])
+ALLOCATION_STATUSES_REQUIRE_END_DATE = ENV.list("ALLOCATION_STATUSES_REQUIRE_END_DATE", default=["Expired"])
+ALLOCATION_STATUSES_DO_DISABLE = ENV.list(
+    "ALLOCATION_STATUSES_DO_DISABLE",
+    default=["Denied", "Revoked"],
+)
+ALLOCATION_STATUSES_DO_UNSET_START_DATE_END_DATE = ENV.list(
+    "ALLOCATION_STATUSES_DO_UNSET_START_DATE_END_DATE",
+    default=["Denied", "New", "Revoked"],
+)
+ALLOCATION_STATUSES_USER_IS_ACTIVE = ENV.list(
+    "ALLOCATION_STATUSES_USER_IS_ACTIVE",
+    default=["Active", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_USER_CHOICES = ENV.list(
+    "ALLOCATION_STATUSES_USER_CHOICES",
+    default=["Approved", "Denied", "Pending"],
+)
+ALLOCATION_STATUSES_DO_REMOVE_USER = ENV.list(
+    "ALLOCATION_STATUSES_DO_REMOVE_USER",
+    default=["Active", "New", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_PAUSRV = ENV.list(
+    "ALLOCATION_STATUSES_PAUSRV",
+    default=["Active", "New", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_ALLOW_REMOVE_USER = ENV.list(
+    "ALLOCATION_STATUSES_ALLOW_REMOVE_USER",
+    default=["Active", "New", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_SHOW_ADD_REMOVE_USER = ENV.list(
+    "ALLOCATION_STATUSES_SHOW_ADD_REMOVE_USER",
+    default=["Active", "New", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_HOMEPAGE = ENV.list(
+    "ALLOCATION_STATUSES_HOMEPAGE",
+    default=["Active", "New", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_AWAITING_ADMIN_ACTION = ENV.list(
+    "ALLOCATION_STATUSES_AWAITING_ADMIN_ACTION",
+    default=["Approved", "New", "Paid", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_SHORT_RENEW_URL = ENV.list(
+    "ALLOCATION_STATUSES_SHORT_RENEW_URL",
+    default=["Payment Pending", "Payment Requested", "Unpaid"],
+)
+ALLOCATION_STATUSES_CAN_EXPIRE = ENV.list(
+    "ALLOCATION_STATUSES_CAN_EXPIRE",
+    default=["Active", "Payment Pending", "Payment Requested", "Unpaid"],
+)
+ALLOCATION_STATUSES_PAYMENT_RELATED = ENV.list(
+    "ALLOCATION_STATUSES_PAYMENT_RELATED",
+    default=["Paid", "Payment Declined", "Payment Pending", "Payment Requested"],
+)
+ALLOCATION_STATUSES_ALLOW_CHANGE = ENV.list(
+    "ALLOCATION_STATUSES_ALLOW_CHANGE",
+    default=["Active", "Paid", "Payment Pending", "Payment Requested", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_COUNT_TOWARDS_LIMIT = ENV.list(
+    "ALLOCATION_STATUSES_COUNT_TOWARDS_LIMIT",
+    default=["Active", "New", "Paid", "Payment Pending", "Payment Requested", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_SUIPBNIA = ENV.list(
+    "ALLOCATION_STATUSES_SUIPBNIA",
+    default=["Active", "New", "Paid", "Payment Pending", "Payment Requested", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_ALLOW_ADD_USER = ENV.list(
+    "ALLOCATION_STATUSES_ALLOW_ADD_USER",
+    default=["Active", "New", "Paid", "Payment Pending", "Payment Requested", "Renewal Requested"],
+)
+ALLOCATION_STATUSES_DO_REMOVE_USER_RENEW = ENV.list(
+    "ALLOCATION_STATUSES_DO_REMOVE_USER_RENEW",
+    default=[
+        "Active",
+        "Denied",
+        "New",
+        "Paid",
+        "Payment Declined",
+        "Payment Pending",
+        "Payment Requested",
+        "Renewal Requested",
+        "Unpaid",
+    ],
+)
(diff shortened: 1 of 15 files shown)


aebruno (changes requested):
Thanks for all the work on this, here's some feedback/thoughts:

- I think this should get merged. I'm not sure we can completely replace statuses with tags. AllocationStatus is an integral part of several core workflows which might be equally challenging replacing them with tags. I really like what you have here and appreciate how thorough you were going through this. It was sorely needed and this really helps clean up the code.
- I like having these as settings and think having an easy way for users to define custom statuses is important. I don't like how these are duplicated though, having to define them in the settings and in a look up table. The same issue would be had if we defined them with constants or emuns as they'd still be duplicated in the lookup table database.
- Instead of defining these as config settings, I'd like to see us try to move towards using [Django Choices](https://docs.djangoproject.com/en/5.2/ref/models/fields/#choices).  Currently we use a lookup table which I think is major source of the pain.
- Choices would be a bit more work as they would require a migration but think it's worth it.
- In addition to allocations, projects and users also have statuses. So if we come up with a generic way of handling choices we could leverage the same code to any model that needs similar functionality.

Here's the basic idea of what I'm proposing. NetBox implements this quite elegantly for example: [here](https://github.com/netbox-community/netbox/blob/main/netbox/dcim/choices.py) and [here](https://github.com/netbox-community/netbox/blob/main/netbox/utilities/choices.py).
1. Change the status field on allocations to be a choice filed
```class Allocation(TimeStampedModel):
    status = models.CharField(
        max_length=50,
        choices=AllocationStatusChoices,
        default=AllocationStatusChoices.STATUS_PENDING
    )
```
2. We define the AllocationStatusChoices to inherit from some base class that implements a basic "choice field" so we can reuse for other objects. This way all status are defined in one place in a nice object that can be used (or even converted to an enum) without comparing  hard coded strings

```class AllocationStatusChoices(ChoiceSet):
    key = 'Allocation.status'
    STATUS_PENDING = 'Pending'
    STATUS_ACTIVE = 'Active'
    STATUS_EXPIRED = 'Expired'

    CHOICES = [
        (STATUS_PENDING, _('Pending')),
        (STATUS_ACTIVE, _('Active')),
        (STATUS_EXPIRED, _('Expired')),
    ]
```
3. The ChoiceSet base class should implement whatever Django requires plus any utility functions, like parsing custom values from settings etc.

```class ChoiceSet():
    def __new__():
        # Here we can parse custom user defined values from settings
    def values(self)
    def as_enum(self)
```
4. Now in the rest of the code and any logic we can do things like:

```if allocation.status == AllocationStatusChoices.STATUS_PENDING:
    # do something
```
5. This will require a migration which would add a new field to the allocations model, set the value to whatever was defined in the current lookup table, then delete the old column and old lookup table.

PR #817, opened 2025-10-05
add missing config

simonLeary42:
`ALLOCATION_RESOURCE_ORDERING` is [documented](https://docs.coldfront.dev/en/stable/config/#coldfront-core-settings) but the code never actually reads for that environment variable.


simonLeary42:
I have learned (https://github.com/ubccr/coldfront/pull/818) that this behavior is shown for a great many settings.

PR #818, opened 2025-10-05
document undocumented options

simonLeary42:
Depending on https://github.com/ubccr/coldfront/issues/815, it may be better to leave `ALLOCATION_FUNCS_ON_EXPIRE` undocumented.


simonLeary42:
new problem: almost all settings and environment variables share the same name, except for `SYSTEM_MONITOR_*` <-> `SYSMON_*`. I can account for this in the regex CI. Do we even want the regex CI?

simonLeary42:
If https://github.com/ubccr/coldfront/pull/853 is merged, then its regex can be removed

simonLeary42:
More readable diff: https://seashells.io/v/Q6WdSzar

PR #828, opened 2025-10-10
use SETTINGS_EXPORT

simonLeary42:
I see two ways of doing the same thing:

`context["FOOBAR"] = import_from_settings("FOOBAR")` -> `{{ FOOBAR }}`

versus:

`SETTINGS_EXPORT.append("FOOBAR")` -> `{{ settings.FOOBAR }}`

I assume that `SETTINGS_EXPORT` is preferred. Doing it the same way everywhere is a good thing, I think.


aebruno:
Thanks for cleaning this up!

simonLeary42:
Found a third way: `{{ settings_value 'FOOBAR' }}`

PR #829, opened 2025-10-12
run tests on all supported python versions, cut 3.9 support

simonLeary42:


aebruno (changes requested):
Let's drop support for python 3.9 as it officially reached EOL as of 2025-10-31. Otherwise, this looks good.

PR #842, opened 2025-10-23
add more to allocation attribute type admin list display

simonLeary42:
before:

image

after:

image


dsajdak:
It was likely not intentional to be left off this display but rather incomplete development.  When these features were added, it was probably not considered to add them to this display but they should be.

PR #849, opened 2025-10-24
add more properties to the allocation type defaults

simonLeary42:
These properties default to `False`, so explicitly setting them to `False` has no effect. I think some of these should be changeable. But in order to set `is_changeable` to `True`, `is_changeable` has to exist in the code.


simonLeary42:
related: https://github.com/ubccr/coldfront/pull/842

aebruno (approved):

PR #850, opened 2025-10-24
project_openldap: check status.name instead of status_id PK

simonLeary42:
comparing integer primary keys is more confusing than comparing string names, and I don't see it adding any benefit.


ds-04:
old method had no benefit, this improves things

PR #854, opened 2025-10-26
dont hide errors

simonLeary42:
You can find many mentions of `non_form_errors` in [the Django docs on formsets](https://docs.djangoproject.com/en/4.2/topics/forms/formsets/).

This is the error I ran into:

> ManagementForm data is missing or has been tampered with. Missing fields: attributeform-TOTAL_FORMS, attributeform-INITIAL_FORMS. You may need to file a bug report if the issue persists.


simonLeary42:
Does it make sense to make a function that does this to remove boilerplate?

PR #855, opened 2025-10-26
add tests, improve tests

simonLeary42:
While writing tests, I discovered that `allocation/views.py` likes to do `message.error` followed by `HttpResponseRedirect`, and the test cases only check `response.status_code`, which is either 200 or 301, depending on if `follow` was specified. I added a new utility to check both `response.status_code` and check for error messages.

* tests tweaked to check for error messages:
    * `test_allocationchangedetailview_post_deny`
    * `test_allocationchangeview_post_extension`
    * `test_allocationchangeview_post_no_change`
    * `test_allocationcreateview_post_zeroquantity`
    * `test_centersummary_renders`
* other test changes:
    * `test_allocationchangedetailview_access` renamed to `test_allocationchangedetailview_access_granted` to differentiate it from `test_allocationchangedetailview_access_denied`
    * `test_allocationchangeview_post_extension`:
        * created local copy of `self.post_data` to prevent this test from messing up other tests with its specific `post_data`
        * added more specific assertion
    * `AllocationViewBaseTest`:
        * added type hints to object attributes
        * added `end_date` to `cls.allocation` so that the new end date could be calculated after approving an end date extension
        * made `cls.quota_attribute.allocation_attribute_type` changeable
    * `test_allocationchangedetailview_post_deny`: tweaked to use `status.name` (string) instead of `status_id` (object primary key), more intuitive
* new tests added:
    * `test_allocationchangedetailview_access_denied`
    * `test_allocationchangedetailview_post_approve_end_date_extension`
    * `test_allocationchangedetailview_post_approve_attribute_change`
    * `test_allocationchangedetailview_post_update_end_date_extension`
    * `test_allocationchangeview_post_attribute_change`

Also added documentation to explain the diffrence between `AllocationChangeView` and `AllocationChangeDetailView`, and their `get_allocation_attributes_to_change`, which share the same name but do different things.

### DIFF
--- a/coldfront/core/allocation/tests/test_views.py
+++ b/coldfront/core/allocation/tests/test_views.py
@@ -3,15 +3,25 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
 import logging
+from datetime import date
 from http import HTTPStatus
 
+from dateutil.relativedelta import relativedelta
 from django.conf import settings
+from django.contrib.auth.models import User
 from django.test import TestCase, override_settings
 from django.urls import reverse
 
 from coldfront.core.allocation.models import (
+    Allocation,
+    AllocationAttribute,
+    AllocationAttributeChangeRequest,
     AllocationChangeRequest,
-    AllocationChangeStatusChoice,
+)
+from coldfront.core.project.models import (
+    Project,
+    ProjectUser,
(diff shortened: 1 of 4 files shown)


simonLeary42:
CI failure is fixed by https://github.com/ubccr/coldfront/pull/847

PR #863, opened 2025-10-31
remove attribute validation boilerplate, add tests, improve tests

simonLeary42:
copied from `Project`:

https://github.com/ubccr/coldfront/blob/46d0ebaa3979d44f37580334afbf7413825c20ed/coldfront/core/project/models.py#L479-L488

I would have removed attribute validation boilerplate from `Resource`, too, but it's incompatible. See https://github.com/ubccr/coldfront/issues/865 and https://github.com/ubccr/coldfront/issues/864.

### DIFF
--- a/coldfront/core/project/tests/test_views.py
+++ b/coldfront/core/project/tests/test_views.py
@@ -39,7 +39,7 @@ class ProjectViewTestBase(TestCase):
         cls.admin_user = UserFactory(is_staff=True, is_superuser=True)
         cls.nonproject_user = UserFactory(is_staff=False, is_superuser=False)
 
-        attributetype = PAttributeTypeFactory(name="string")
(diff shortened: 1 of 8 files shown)


coldfront/core/project/tests/test_views.py
```diff
@@ -39,7 +39,7 @@ def setUpTestData(cls):
         cls.admin_user = UserFactory(is_staff=True, is_superuser=True)
         cls.nonproject_user = UserFactory(is_staff=False, is_superuser=False)
 
-        attributetype = PAttributeTypeFactory(name="string")
+        attributetype = PAttributeTypeFactory(name="Text")
```
simonLeary42: `string` is not a valid attribute type.

Issue #848, opened 2025-10-24
ColdFront Database Query Amount Analysis


matthew-kusz:
While working on coldfront I noticed it is producing a large number of database queries. I've started an analysis on where these queries are coming from and why there are so many and was wondering if anyone else has done something similar? I've been going through the code and looking for ways to reduce the number of queries, especially for pages that receive frequent traffic. I can post my findings here once I'm done, or in a PR with the changes.

To give a couple of examples, I set up a fresh install of coldfront and loaded in the test data. I used the admin account and created a total of 6 projects and 6 allocations and right now have only been testing with sqlite. With this the home page had 57 database queries that I was able to reduce down to 29, and the allocation summary on the center summary page had 47 queries that I was able to reduce down to 19.

PR #875, opened 2025-11-12
Database Query Reductions

matthew-kusz:
This PR involves changes to various places on the site that queries databases. When testing this out I was using a SQLite database on a fresh ColdFront install, but none of the changes should affect other databases. The motivation for this was I noticed a large number of queries ran on the home page and I decided to investigate other pages, #848. I mostly stuck to easy changes that shouldn't change the code too much. There's likely ways to rework some of the logic that cause additional database queries but I think this is a good start. One function that kept coming up that causes additional queries was the `get_parent_resource` function in the `Allocation` class, and I haven't found an easy way to handle that yet. The biggest offenders were pages that listed project, allocations, resources, etc. Some of the database queries on those pages scaled with the number of entries.

If a table of the improvements on each page is needed I can provide that once I have it finished, I've been keeping track of improvements on each individual page but haven't set up a proper table with my findings.

### DIFF
--- a/coldfront/core/allocation/forms.py
+++ b/coldfront/core/allocation/forms.py
@@ -133,7 +133,7 @@ class AllocationSearchForm(forms.Form):
     )
     resource_name = forms.ModelMultipleChoiceField(
         label="Resource Name",
-        queryset=Resource.objects.filter(is_allocatable=True).order_by(Lower("name")),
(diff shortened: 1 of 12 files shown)


matthew-kusz:
It is ready to review, I was referring to including the table if it's needed.

PR #904, opened 2025-12-08
Upgrade to Datatables 2

cecilialau6776:


aebruno:
> My latest commit should have that fixed... 

Ah yep, that fixed it. I had a custom css that didn't have this change. All looks good.

PR #907, opened 2025-12-09
Batch commands when syncing FreeIPA.

aebruno:
When syncing unix groups with FreeIPA, the group_remove_member and group_add_member commands are submitted to the IPA server separately per user. This patch modifies the FreeIPA sync command to leverage the batch command from the IPA API which can be used to send commands in the same API request. Currently the batch size is hard coded to 100 which should be sufficent. In the future consider making this a CLI arg.

See here: https://freeipa.readthedocs.io/en/latest/api/basic_usage.html#batch-operations

We have also been experiencing random failures on IPA API calls, for example:

```
ipa: ERROR: connect to 'https://ipa-server/ipa/json': EOF occurred in violation of protocol (_ssl.c:2427)
```

It appears this may be caused by running the IPA API bootstrapping when utils.py was imported instead of right before making an API call.  This patch adds a helper function for more control over when to bootstrap the IPA API.

### DIFF
--- a/coldfront/plugins/freeipa/management/commands/freeipa_check.py
+++ b/coldfront/plugins/freeipa/management/commands/freeipa_check.py
@@ -18,9 +18,7 @@ from coldfront.plugins.freeipa.utils import (
     CLIENT_KTNAME,
     FREEIPA_NOOP,
     UNIX_GROUP_ATTRIBUTE_NAME,
-    AlreadyMemberError,
-    NotMemberError,
-    check_ipa_group_error,
+    ipa_bootstrap,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


coldfront/plugins/freeipa/management/commands/freeipa_check.py
```diff
@@ -76,16 +65,7 @@ def add_group(self, user, group, status):
...
-            try:
-                res = api.Command.group_remove_member(group, user=[user.username])
-                check_ipa_group_error(res)
-            except NotMemberError:
-                logger.warning("User %s is not a member of group %s", user.username, group)
-            except Exception as e:
-                logger.error("Failed removing user %s from group %s: %s", user.username, group, e)
-            else:
-                logger.info("Removed user %s from group %s successfully", user.username, group)
+        self.ipa_batch_args.append({"method": "group_remove_member", "params": [[group], {"user": [user.username]}]})
```
simonLeary42: You could make this more elegant and avoid a breaking API change if you made these functions take a `str|list[str]` username or list of usernames, rather than changing these functions from write-through (changes are applied immediately) to write-back (changes are buffered and you must remember to apply them).

PR #913, opened 2025-12-16
Allow project managers to view allocations associated with the project

s-sajid-ali:
## Description

Per discussion in the recent Coldfront community conversation, this PR allows a Manager for a project to view the Allocations associated with the project. Note that this allows a manager to also edit the allocations since anyone with `ProjectPermisson.MANAGER` is implicitly assigned `AllocationPermission.MANAGER`.

## Component

- [x] Projects
- [x] Allocations

### DIFF
--- a/coldfront/core/project/tests/test_views.py
+++ b/coldfront/core/project/tests/test_views.py
@@ -9,6 +9,10 @@ from django.test import TestCase
 from coldfront.core.project.models import ProjectUserStatusChoice
 from coldfront.core.test_helpers import utils
 from coldfront.core.test_helpers.factories import (
+    AllocationFactory,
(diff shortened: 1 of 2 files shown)


dsajdak:
LGTM!  I would just clarify that this allows a manager on a project to view all allocations on the project, even if the manager isn't a user on the allocation.  Currently, despite being a manager on the project, if the user isn't an active allocation user they can't view that allocation.

s-sajid-ali:
Correct, that is the intent, i.e. to allow project managers to add/remove members to the allocations as well as the project.

s-sajid-ali:
Added a test as that fails without the patch.

PR #918, opened 2025-12-23
Fixes #917.

### LINKED ISSUE #917: Bug: gauge chart shows used plus total instead of used plus available
### What happened?

The gauge should show the used amount plus the remaining amount, but shows the used amount plus the total amount. 

Image

in gaugeChart.ts, I think you can just change the data in the createGaugeChart function to:
data: [value, (total - value)]

to get the expected output

### Version

latest

### Component

_No response_

None

### Relevant log output

```shell

```

- [ ]

aebruno:
Gauge chart should show the used amount plus the remaining amount.


aebruno:
Fixes #917

PR #944, opened 2026-02-12
remove python 3.10, 3.11 from CI

simonLeary42:


simonLeary42:
I brought this up in the last core arch workgroup meeting, and the justification given was that `uv` manages its own python, so there's no reason to be bound by the system python version of various linux distros.

PR #747, opened 2025-07-30
Fixed formatting in loggers.

Eric-Butcher:
Address most of the remaining concerns in #709. Adds new ruff check rules for log formatting to make future PR reviews go smoother.

Issue #709, opened 2025-06-26
Fix the way Coldfront handles logging


Eric-Butcher:
The way Coldfront handles logging should be improved. 

There are two issues that I have encountered with the way Coldfront handles logging. 

The first issue is the usage of deprecated logging methods. There are multiple instances throughout coldfront where `logging.warn()` is used. This method has been deprecated in favor of `logging.warning`. This deprecated method was fully removed in Python 3.13 which would mean that any user upgrading their coldfront deployment to run on python 3.13 or greater could receive a runtime error if they ever encounter one of these statements. There are also instances where the level `logging.WARN` is used which should instead by `logging.WARNING` to reflect the modern logging naming. See [Python's logging documentation](https://docs.python.org/3.12/library/logging.html).

The second issue is the improper usage of custom formatting strings when used in logging. It is a best practice to use the integrated formatting functionality of python's logging module instead of formatting strings manually outside of the logger.  This is for reasons including:
1. when using internal logic, the string will never be created if your log level is set higher than the logging statement
2. if an exception or error is created when trying to create the log string with the internal log, the log still gets created saying it failed to create the string instead of crashing and never producing the log of what went wrong
3. the log format is generally faster

All the problems noted with logging and more can be automatically detected by configuring ruff to check for:
- "LOG": https://docs.astral.sh/ruff/rules/#flake8-logging-log
- "G": https://docs.astral.sh/ruff/rules/#flake8-logging-format-g
- "PLE1205, PLE1206": https://docs.astral.sh/ruff/rules/#convention-plc
- "TRY400, TRY401": https://docs.astral.sh/ruff/rules/#tryceratops-try

Recommendations:
1. Do a coldfront wide find and replace refactor for deprecated logging functionality
2. Do a coldfront wide refactor of logging string formatting
3. Configure ruff linter to check for coming guideline violations when using logging

PR #976, opened 2026-03-31
Log errors where applicable

cecilialau6776:
Makes sure applicable errors get logged. I ran into this problem when sending mail and did not get a detailed error.


aebruno:
@cecilialau6776 This looks good and is related to #709 and #747. Can you fix up this PR so that it conforms to the suggestions in #709? If we do end up merging those then this code will not have to be changed. Thanks!

aebruno (changes requested):
See #709

cecilialau6776:
Yup! I didn't know about those. Any way I can help with getting #709 merged as well?

PR #981, opened 2026-04-02
project_openldap: fix backwards conditional

simonLeary42:


simonLeary42:
https://github.com/coldfront/coldfront/pull/983

PR #982, opened 2026-04-02
project_openldap: fix entries

simonLeary42:
after `Connection.unbind()`, `entries` is overwritten.


simonLeary42:
https://github.com/coldfront/coldfront/pull/983

PR #991, opened 2026-04-14
Add support for subpaths

### LINKED ISSUE #988: Bug: Cannot run Coldfront under subpath
### What happened?

I want to run Coldfront under a subpath (such as `http://example.com/coldfront`). I cannot get this to work for static files because Coldfront does not add the `/coldfront` bit in front of any paths to static files referenced in the HTML returned from gunicorn.

I have tried various incarnations of the [`SCRIPT_NAME` header](http://albertoconnor.ca/hosting-django-under-different-locations.html) and the [`FORCE_SCRIPT_NAME` config option](https://docs.djangoproject.com/en/5.2/ref/settings/#force-script-name), but I could not get any of them work.

Is there something I'm missing here? Has anyone else successfully deployed Coldfront on a subpath?

### Version

1.1.7

### Component

Other

Other

### Relevant log output

```shell

```

1. Deploy Coldfront on subpath (e.g. `/coldfront`)
2. `curl example.com/coldfront`
3. Check if links to static files have the `/coldfront` in their paths

runiq:
This PR is supposed to add support for running Coldfront on a subpath using `FORCE_SCRIPT_NAME`, which we need for our center. There are two things still missing, that's why this is marked as draft: Docs, and Vite-generated assets.

Currently the `coldfront.css` is checked into the repo with static paths, but I would need to recreate these files with the `FORCE_SCRIPT_NAME` in front, probably on `collectstatic`. Unfortunately I have no experience with Vite; I would love it if anyone could point me into the right direction on how to wire this up.

Fixes https://github.com/coldfront/coldfront/issues/988.

### DIFF
--- a/coldfront/config/auth.py
+++ b/coldfront/config/auth.py
@@ -13,7 +13,7 @@ AUTHENTICATION_BACKENDS += [
 ]
 
 LOGIN_URL = "/user/login"
-LOGIN_REDIRECT_URL = "/"
+LOGIN_REDIRECT_URL = "home"
 LOGOUT_REDIRECT_URL = ENV.str("LOGOUT_REDIRECT_URL", LOGIN_URL)
 
 SU_LOGIN_CALLBACK = "coldfront.core.utils.common.su_login_callback"
--- a/coldfront/config/auth.py
+++ b/coldfront/config/auth.py
@@ -12,12 +12,12 @@ AUTHENTICATION_BACKENDS += [
     "django.contrib.auth.backends.ModelBackend",
 ]
 
-LOGIN_URL = "/user/login"
+LOGIN_URL = "user/login"
 LOGIN_REDIRECT_URL = "home"
 LOGOUT_REDIRECT_URL = ENV.str("LOGOUT_REDIRECT_URL", LOGIN_URL)
 
 SU_LOGIN_CALLBACK = "coldfront.core.utils.common.su_login_callback"
-SU_LOGOUT_REDIRECT_URL = "/admin/auth/user/"
+SU_LOGOUT_REDIRECT_URL = "admin/auth/user/"
 
 SESSION_COOKIE_AGE = ENV.int("SESSION_INACTIVITY_TIMEOUT", default=60 * 60)
 SESSION_SAVE_EVERY_REQUEST = True
--- a/coldfront/config/base.py
+++ b/coldfront/config/base.py
@@ -149,7 +149,7 @@ CRISPY_TEMPLATE_PACK = "bootstrap5"
 
 SETTINGS_EXPORT = []
 
-STATIC_URL = "/static/"
+STATIC_URL = "static/"
 
 DJANGO_VITE = {
     "default": {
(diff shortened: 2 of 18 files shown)


aebruno:
@runiq Thanks for the PR. 

> Currently the `coldfront.css` is checked into the repo with static paths, but I would need to recreate these files with the `FORCE_SCRIPT_NAME` in front, probably on `collectstatic`. Unfortunately I have no experience with Vite; I would love it if anyone could point me into the right direction on how to wire this up.

[This line](https://github.com/coldfront/coldfront/blob/main/coldfront/static/vite.config.js#L24) in the `vite.config.js` has the hard coded path that will need adjusting. This will likely just need to be part of the instructions for adding a base path. If you set a custom base path, you'll need to rebuild the static assets. Instructions for building can be found [here](https://github.com/coldfront/coldfront/blob/main/coldfront/static/README.md). We could make this an env variable like we do for `env.APP_PORT` but the assets will still need to be re-compiled.

runiq:
Hmm, that's unfortunate :( I was hoping there'd be a way to incorporate regenerating the assets into `collectstatic` or something similar.

So basically, when I host Coldfront on a subpath, I have to always regenerate the assets and there's no way to do so dynamically? If so, I'll keep that in mind for the documentation.

aebruno:
> Hmm, that's unfortunate :( I was hoping there'd be a way to incorporate regenerating the assets into `collectstatic` or something similar.
> 
> So basically, when I host Coldfront on a subpath, I have to always regenerate the assets and there's no way to do so dynamically? If so, I'll keep that in mind for the documentation.

@runiq Did some more testing on this and looks like this is an easy fix. Should be able to use relative paths here. If you want to test this out you can set `base: ""` to an empy string in `vite.config.js` and rebuild the static assets. [see this patch](https://github.com/coldfront/coldfront/pull/972/changes#diff-b55558ba9f58220950b414c8e2dfa0c1e33db5ac50c16a90f8a984fc9f3a5bcd). I'll be making this change permanent after some more testing so will be included by default in future releases. This should enable the included static assets to be served from any sub-path.

runiq:
That would be amazing! Unfortunately I'm about to head home for the day; I'll test tomorrow and report back.

Thanks for looking into this!

runiq:
I tested your proposal and it appears to work. No more wrong subpath in Firefox devtools. (I was too lazy to add npm to my container, so I manually adjusted the paths in the minified .css and .js files instead.)

I also fixed some things I missed the first time. Modulo docs, I think this is good enough for testing. I'll write docs tomorrow.

aebruno:
@runiq Thanks again for your input, I refactored this a bit in cdc2eb6caaa9a41a2c5849af48bf20a16f3466b6 which is now merged and support for serving ColdFront from a sub-path (base path) is included in v1.1.8. You can now just set the `BASE_PATH=/mysubpath` config option.

runiq:
Thank you for pushing this to the finish line! :pray: Sorry, I got sidetracked by other commitments.

PR #998, opened 2026-05-07
Update MoneyField and remove PercentField

### LINKED ISSUE #888: address concerns with percent field, money field
some changes were discussed in https://github.com/coldfront/coldfront/pull/834, but left out of scope:

https://github.com/coldfront/coldfront/pull/834#issuecomment-3527901565

https://github.com/coldfront/coldfront/pull/834#issuecomment-3528080591

https://github.com/coldfront/coldfront/pull/834#issuecomment-3528289415

cecilialau6776:
Adds a new setting `DEFAULT_CURRENCY`, which is currently set to `"USD"`

Requires a migration, which sets the default currency to `DEFAULT_CURRENCY`. The migration will pull from the setting, so if you want the default currency to be something other than USD, change the setting before running the migration. _Any existing grants will have their currency updated to this default._

Adds the `djmoney` package.

Comments:
The grant create and update pages have the form split up from `{{ form | crispy }}` into separate fields so the percentage field can have a "%" appended to it. I changed this in the HTML template rather than using a `cripsy` layout because our center hides some of the fields this way.

Should close #888.


aebruno:
@cecilialau6776 apologies for the delay on reviewing this. Can you rebase/re-submit this against the `stable/1.1.x` branch? Thanks!

PR #1031, opened 2026-09-08
add tests for project_openldap

simonLeary42:
* also made it so the slurm plugin tests will now run in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,8 +30,11 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      - name: install system packages
+        run: DEBIAN_FRONTEND=noninteractive sudo apt install -y libldap2-dev libsasl2-dev
+
       - name: Install the project
-        run: uv sync --locked --dev
+        run: uv sync --locked --dev --extra ldap
 
       - name: Install frontend packages
         run: npm --prefix coldfront/static install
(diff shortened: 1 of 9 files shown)


coldfront/plugins/project_openldap/tasks.py
ds-04, line 1: This is done because when the project create signal fires, only the PI can be known.

.github/workflows/ci.yml
```diff
@@ -62,7 +65,7 @@ jobs:
         run: uv run reuse lint
 
       - name: Run tests
-        run: uv run coldfront test
+        run: PLUGIN_SLURM=true PLUGIN_PROJECT_OPENLDAP=true PROJECT_OPENLDAP_GID_START=-1 uv run coldfront test
```
aebruno, line 68: This a plugin so let's not run these as part of the default CI. Feel free to add some docs to a readme that describes how to run the tests for this plugin.

aebruno (changes requested):

.github/workflows/ci.yml
simonLeary42: But how will we know if someone (me) opens a PR with a regression?

simonLeary42: What if I make it so it runs the tests if it detects that one of the plugin's files has changed?

coldfront/plugins/project_openldap/tasks.py
simonLeary42: Updated to clear up the confusion:
* removed `_create_project_with_users` function because it's unnatural for a project to already have users before `tasks.add_project()` runs
    * instead, explicitly call `ProjectFactory()`, then `tasks.add_project()`, then `ProjectUserFactory()`, then `tasks.add_user_project()`
* added comments and print statements to the mgmt command to document the quirk that the command needs to be run a second time to fully sync
* added comment to `tasks.add_project` about why it doesn't need to look for users

